PR #8, opened 2025-01-31
Feat: adds ability to navigate through user's working directory

allison-truhlar:
- Mimicking the JupyterLab interface, double-clicking on a directory navigates into the directory. Single-clicking just checks/unchecks the directory.
- Includes breadcrumbs at the top of the file list to navigate back to higher directory levels.

Note that no functionality exists when double-clicking on a file. This will be addressed through issue #7.


allison-truhlar:
@krokicki @neomorphic

PR #12, opened 2025-02-14
fix: adds pydantic as a test depedency

allison-truhlar:
pydantic is used by the filestore. it is installed in the pixi environment through copier, but it does not get automatically installed in the github action test environment, so we must make that explicit


allison-truhlar:
Note - the Binder link can sometimes fail if binder.org is down/lagging. Check [here](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub) for outages.

PR #9, opened 2025-02-11
Filestore API

krokicki:
Added a Filestore API for listing files with their owner/groups/permission information, and other things that the built-in Contents API does not provide.

PR #13, opened 2025-02-20
Feat: enable routing in widget via react-router

allison-truhlar:
@krokicki @neomorphic This PR adds `react`, `react-dom`, and `react-router` as dependencies in `package.json`. Then, in `src/App.tsx`, I set up a very simple `BrowserRouter` for now, just to demo the routing.

Finally, unrelated to the above, I removed the `skipLibCheck:true` flag from `tsconfig.json` and `tsconfig.test.json`, since this was resolved by pinning a `jest` dependency in the `resolutions` section of `package.json`. See further conversation [here](https://github.com/JaneliaSciComp/fileglancer/pull/9#issuecomment-2659833084).

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,10 @@
         "@jupyterlab/services": "^7.0.0",
         "@jupyterlab/ui-components": "^4.3.4",
         "@mui/icons-material": "^6.4.1",
-        "@mui/material": "^6.4.1"
+        "@mui/material": "^6.4.1",
+        "react": "18.3.0",
+        "react-dom": "18.3.0",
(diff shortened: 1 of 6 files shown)


allison-truhlar:
@neomorphic Example URL with additional segments after `lab/`: http://localhost:8888/lab/workspaces/auto-6
I think this happens when you have an old JupyterLab tab open and then start up JupyterLab again from the console. For example, when I didn't close that tab and then launched JupyterLab again in the console, the next tab URL was: http://localhost:8888/lab/workspaces/auto-H

PR #15, opened 2025-02-28
Fix install

krokicki:
Minor fixes I needed to make to install this on a new laptop. 

@StephanPreibisch


allison-truhlar (approved):
Thanks for fixing the commands and adding `npm-run-all`!

PR #17, opened 2025-03-03
Adds navbar across pages; converts FileList to material-tailwind

allison-truhlar:
@krokicki @neomorphic 
This PR utilizes the Tailwind and React-Router dependencies that were added in previous PRs to add a navbar component that appears across the `lab/` (index), `lab/files`, `lab/jobs`, and `lab/help` routes. Note: the `lab/` prefix is added by JupyterLab - this may need to change with JupyterHub. I settled on [material-tailwind](https://www.material-tailwind.com/) for the components after also trying materialUI and shadcn. MaterialUI works well out of the box but doesn't integrate well with Tailwind styling; shadcn works well with Tailwind but requires more composition of small components to build blocks like a navbar, and the styling was ultimately more difficult to predictably apply because of this. 

Because of the switch to material-tailwind from the original usage of materialUI in the FileList component, this PR also refactors that component to use material-tailwind. I also separated clicking on the checkboxes in the FIleList from clicking on the filename. Previously, clicking once checked the checkbox and clicking twice changed what was displayed in the file list. Now, a single click on the checkbox selects the checkbox, and a single click on the file or directory name changes what's displayed in the file list.


allison-truhlar:
I was just looking this over again after merging Konrad's file share changes and there are a few small things I'd like to add before merging:

- [ ] Links for the profile and preference pages in the profile avatar dropdown
- [ ] Remove the rounded top corners of the navbar

PR #20, opened 2025-03-11
chore: Updates pyproject.toml to build on linux-64

neomorphic:
- Added "linux-64" to platforms.
- Also had to downgrade the nodejs version as 22.13.0 is not available
  for linux.


krokicki:
LGTM

krokicki (approved):

PR #27, opened 2025-03-20
Build refactoring

krokicki:
@allison-truhlar @neomorphic @StephanPreibisch 

I kept running into weird build issues when installing on new machines or after switching branches, so I took some time to attempt to understand this complex build from first principles, and then refactored the build process to attempt to make it more predictable. Feedback would be greatly appreciated! 

## Overview 

There are three main components to the build (configured in `pyproject.toml`):
1) The Pixi environment (containing Node, the Python interpreter, and any Python dependencies which may be from PyPI or Conda) 
2) The Python-based backend extension using *hatch* (configured as the build system in `pyproject.toml`)
3) The Node-based frontend extension using *npm* (configured in `package.json`)

For development builds, there are several independent steps:
1) Initial dependency install and environment set up - only needs to be rerun when dependencies change
2) Watch for changes to the frontend code and recompile
3) Run JupyterHub, watching for changes to the Python code

## Challenges

The build is complex, has many moving parts, and includes several build systems which can interfere with each other if not properly managed. Unfortunately, certain dependencies generate code outside of a build directory, making it more difficult to clean the environment. With an unclean build environment, I would sometimes run into errors like:
```
sh: line 1: run-p: command not found
```
I believe that this was due to steps running out of order. In particular, for the above error, `npm install` needs to run before `npm run watch`. 

Another recurring error was:
```
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory build_log.json
```
This was caused by the file existing (contrary to the error message), and resolved when the file is deleted. 

One more I couldn't explain was this:
```
The package `pytest-jupyter "does not have an extra named `server`"
```
I'm guessing this has to do with conflicts in Python packages being installed with both PyPI and Conda. 

### Clean Step

I started by creating a new cleaning script called `clean.sh`. This script deletes the large `node_modules` and `.pixi/envs` folders in the background, so you don't need to wait for the files to be removed before proceeding. It also deletes any other generated code which can (and does) cause issues in certain situations. As far as I can tell, this clean procedure cannot be a Pixi task because if you run it on a clean dev environment, Pixi will rebuild the environment before it trashes it. 

### Pixi Environment

To build the Pixi environment, I ordered the necessary steps as follows:

```mermaid
flowchart TD
    A[ext-clean] --> B
    B[pip-install] --> F
    C[node-install] --> F
    D[ext-overwrite] --> E
    E[ext-enable] --> F
    F[dev-install]
```

This is a one-time process which installs the package in editable mode using pip (after cleaning labextensions/fileglancer, which otherwise causes the strange "No such file or directory build_log.json" described above). Then it runs the NPM dependency install, and finally activates the JupyterHub extension. I believe all these steps only need to be done once, but it's possible I overlooked something. 

### Frontend Rebuild

I renamed this step from `dev-build` to `dev-watch` to be more obvious about what it does. It no longer depends on anything since the extension installation was moved to the `dev-install` step.

### JupyterHub 

`dev-launch` has not changed, but I added a `dev-launch-remote` for running on a remote server.

## Other Changes

I made a few other changes as I was getting the build working, which I believe make things smoother (but again, open to suggestions). 

1) Moved all Python dependencies from `tool.pixi.dependencies` (Conda) to `dependencies` (PyPI) and removed some which did not seem necessary (they may end up being necessary but I'd like to understand why). The only dependencies being installed with Conda are python, pip, and nodejs. I think it is better to keep as much as possible in a single build system, and this seems to have eliminated the pytest-jupyter issue described above.
2) Removed both lock files from GitHub. These are constantly changing and I'm not sure if there is a benefit to version controlling them, at least early on while things are still changing rapidly. 
3) Added `test-backend` and `test-frontend` tasks to run the tests inside the "test" environment.
4) Removed `npm-run-all` and left `npm-run-all2`. Since one is a fork of the other I think we only need one of these.
5) Added `junit.xml` to `.gitignore` since it is a test output.
6) Require Python >=3.10. We should increase this as much as we can so that we're not supporting old versions. 
7) Updated the documentation to reflect all the build changes.

## Unaddressed Issues

* I haven't considered the release build at all. 
* There is a generated file `style/tailwind_converted.css` which is checked into Git, but maybe shouldn't be?
* Whenever the pyproject.toml changes, Pixi decides that the entire environment needs to be rebuilt, even if you are just tweaking a test task. It seems like there should be a way to exclude an editable package from the lock file. 
* Pixi seems to cross compile for all the included platforms, even when they are not needed. It would be nice to have a way to build only for the current platform. 
* Removing the pydantic test dependency has broken the automated testing. I think it needs to run in Pixi for it to work.


.gitignore
```diff
@@ -127,3 +128,8 @@ dmypy.json
 # pixi environments
 .pixi
 *.egg-info
+
+# lock files
+pixi.lock
+package-lock.json
```
neomorphic, line 134: I would like to have these in version control at some point. Not having them can lead to a "works on my machine" scenario, which is the whole reason for their existence.

neomorphic (approved):
It all seems reasonable to me. I did note that I would like to get the lock files back under version control eventually.

allison-truhlar:
I fixed the GH action errors, but I'm not fully understanding how the GH action is initiating the build. It seems that `pydantic` and `jupyterlab` are now installed inside `fileglancer.whl`, which was what was throwing the first error in the automated testing (see below).  I guess as a result of Konrad moving them up into the `dependencies` section of `pyproject.toml`. But I don't see where in `build.yml` that `pyproject.toml` is being invoked.

The other thing that needed to be changed in the GH action was updating the installed version of Python (previously 3.9) to match the higher version requirements Konrad added.

```
 INFO: pip is looking at multiple versions of fileglancer to determine which version is compatible with other requirements. This could take a while.
Collecting pydantic
  Downloading pydantic-2.10.5-py3-none-any.whl.metadata (30 kB)
ERROR: Cannot install fileglancer==0.1.0 and pydantic because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pydantic
    fileglancer 0.1.0 depends on pydantic<3 and >=2.10.6

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict
```

allison-truhlar (approved):
Looks good to me. I made a couple of changes to fix the broken GH action. I also added `tailwind_converted.css` to the `.gitignore`, as suggested, since this file gets regenerated on every build step.

PR #35, opened 2025-04-07
Fix breadcrumbs

allison-truhlar:
@krokicki @neomorphic 
A small fix I forgot to include in my file list UI improvement PR last week. The breadcrumbs now correctly include the file zone name as a breadcrumb link in addition to the file path segment links. Further, following Starfish, the squares icon no longer functions as the link for the file zone.


allison-truhlar:
Note that the `build` GH action is failing at command `python -m jupyterlab.browser_check`, but I think this has to do with how `jupyter-server` is loading the app, so I decided not to keep working on fixing the build since we're prioritizing the standalone app verison.

PR #37, opened 2025-04-11
Add favoriting and path preference functionality

allison-truhlar:
@neomorphic @krokicki 

The two main purposes of this PR are: 
1. Add functionality for the user to select zone, file share path, and directory favorites, which are then displayed in a favorites component in the sidebar. It closes this [clickup ticket](https://app.clickup.com/t/86a76kk5y).

  - Clicking on a star next to a zone or file share path toggles the zone or file share path as a favorite.
  - Right clicking on a folder (or folders) and then selecting "set/unset favorite" from the context menu toggles directories as favorites. You can also click the "..." button in the directory row in the file browser to access the context menu and add the directory as a favorite.
  - Once favorited, clicking on: 
    - a zone opens the zone in the zone browser
    - a file share path opens the corresponding zone in the zone browser and the file share path in the file browser
    - a directory opens the corresponding zone in the zone browser and the directory in the file browser

2. Add a preferences page to allow the user to select their path display preferences. Closes this [clickup ticket](https://app.clickup.com/t/86a7j9xde)

The PR also includes a major refactor of state, context, and hooks to manage the state and context.

It also adds the new Fileglancer SVG icon with a blue/green to purple gradient to the JupyterLab launcher.

PR #38, opened 2025-04-21
Fix file browser

allison-truhlar:
@krokicki @neomorphic 
This fixes the bug introduced in browsing the file list by PR #37. I added new state, so now the current zone (used only for the sidebar zones UI), current file share path (used for fetching files), and current full navigation path (used only for the file crumb list UI) are stored in state.
You can test the functionality at https://fileglancer-dev.int.janelia.org, where this branch is currently installed.


krokicki (approved):
LGTM

PR #41, opened 2025-05-05
Toolbar additions and some misc fixes

allison-truhlar:
@neomorphic @krokicki 
This PR addresses this [clickup ticket](https://app.clickup.com/t/86a84r9r3).
- Commits 941c24c9dabe972b6adc9dbf95f471840fe3afbd and 3e131390ce4b08584996215d293cfd8137267725: add the button on the toolbar to open/close the file properties drawer
- Commits 08287adcc0f60f6fe24a45dc6cab8d418ea492ac through 46c2377127621eac09be9d071a26468316272e7d: add a "new folder" button to the file properties drawer

This PR also contains two fixes unrelated to the above, that I noticed in the course of work:
- Commit fa2ce8d81cef0f85056f49503dfa6cae005a7542: allows multiple selected directories to be added as favorites at once
- Commit 30c255ad6bcf3ed62532182d86411c5d609ff716: there was a change in the format of how file share path information was sent to the file browser crumbs (it's now the full file share path object, not just the path name), so the name needed to be accessed on the object


src/contexts/PreferencesContext.tsx
```diff
@@ -147,15 +147,29 @@ export const PreferencesProvider = ({
...
+      existingItemIndex.forEach((itemIndex, index) => {
+        if (itemIndex >= 0) {
+          newFavorites.splice(itemIndex, 1);
+        } else {
+          newFavorites = [...newFavorites, newItem[index]];
+        }
+      });
+    } else if (
+      typeof existingItemIndex === 'number' &&
+      !Array.isArray(newItem)
```
krokicki, line 164: This if/else is dangerous because it does nothing in several cases. I think it would be better to have each if condition check a single condition (e.g. `if (Array.isArray(existingItemIndex)`) and then have nested ifs for multiple conditions. That would allow us to handle every else case and throw errors for invalid cases. 

But before making any additional changes, let's discuss how these preferences are being handled at a conceptual level, and if the scalar/array duality is really necessary. It would be ideal to delegate this kind of complexity to the backend if possible.

src/components/ui/FileBrowser/Crumbs.tsx
```diff
@@ -37,11 +37,13 @@ export default function FileListCrumbs(): JSX.Element {
...
                 onClick={() => {
-                  if (index === 0) {
-                    fetchAndFormatFilesForDisplay(`${currentFileSharePath}`);
-                  } else {
+                  if (index === 0 && currentFileSharePath) {
                     fetchAndFormatFilesForDisplay(
-                      `${currentFileSharePath}?subpath=${dirArray.slice(1, index + 1).join('/')}`
+                      `${currentFileSharePath.name}`
+                    );
+                  } else if (currentFileSharePath) {
```
krokicki, line 44: Looks like this could be extracted to an enclosing if statement, something like this:
```
if (currentFileSharePath) {
  if ( index === 0 ){
  } 
  else {
  }
}
```

src/contexts/PreferencesContext.tsx
```diff
@@ -189,20 +203,36 @@ export const PreferencesProvider = ({
...
 
-  function handleDirectoryFavoriteChange(item: DirectoryFavorite) {
-    const existingItemIndex = directoryFavorites.findIndex(
-      dir => dir.name === item.name && dir.fileSharePath === item.fileSharePath
-    );
+  function handleDirectoryFavoriteChange(
+    item: DirectoryFavorite | DirectoryFavorite[]
+  ) {
+    let existingItemIndex;
+    if (Array.isArray(item)) {
```
krokicki, line 210: This might be okay, but we want to try to avoid type checks whenever possible, which are generally considered to be [code smells](https://dev.to/mcsee/code-smell-23-instance-type-checking-1i90). They can usually be replaced with polymorphism or some kind of data modeling.

krokicki (changes requested):
Hi @allison-truhlar, 
The deployed site looks great, I just have some questions about the code that we could discuss in person. 
I was thinking that another button we could use in the toolbar is a refresh button. Since the normal browser refresh doesn't work, there's no way to get an updated file list when the file system changes outside of Fileglancer.

allison-truhlar:
@krokicki I will make the preferences changes you requested and we discussed to another PR.

PR #43, opened 2025-05-09
Update preference management

allison-truhlar:
@krokicki @neomorphic 
This PR is now ready for review. I can wait to fully merge it until I integrate the build changes from Jody.

The main change is adding the construction of the "uber map" in `FileBrowserContext.tsx` and updating the preference management to store hash keys. The format for these keys is the same as in the "uber map" for zones (`zone_{fsp.zone}`) and FSPs (`fsp_{fsp.name}`). The folder preference key format is `folder_{fsp.name}_{folder.path}`. In addition to being used to access zones and fsps for the favorites, the "uber map" is now also used for populating the `ZoneBrowser`.

I also updated some types/types names in `shared.types.ts`, most notably `File` to `FileOrFolder`, since that's a more accurate description and also avoids the error that would sometimes occur if I forgot to import the `File` type, and the built-in JS interface `File` would be assigned as the type.


src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
-      // Sort the items within each zone alphabetically by name
-      Object.keys(unsortedPaths).forEach(zone => {
-        unsortedPaths[zone].sort((a, b) => a.name.localeCompare(b.name));
+  async function updateZonesAndFileSharePathsMap() {
+    let rawData: { paths: FileSharePath[] } = { paths: [] };
+    try {
+      rawData = await getZones();
+      const newZonesAndFileSharePathsMap: Record<string, FileSharePath | Zone> =
+        {};
+      rawData.paths.forEach(item => {
```
krokicki, line 107: I would move this for-loop logic along with line 105 into `createZonesAndFileSharePathsMap`, or just make these a single function. Splitting the functionality here makes it harder to reason about the purpose and functionality of`createZonesAndFileSharePathsMap`.

src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
+    // If the zone doesn't exist in the map, create it
+    if (!newZonesAndFileSharePathsMap[`zone_${item.zone}`]) {
+      newZonesAndFileSharePathsMap[`zone_${item.zone}`] = {
+        name: item.zone,
+        fileSharePaths: []
+      } as Zone;
+    }
+    // If/once zone exists, check if file share path exists in the zone
+    // And update zone accordingly
+    if (newZonesAndFileSharePathsMap[`zone_${item.zone}`]) {
```
krokicki, line 81: Unless I'm misunderstanding something, this check should not be necessary, because the key is guaranteed to exist due to the code above.

src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
-        if (!unsortedPaths[item.zone]) {
-          unsortedPaths[item.zone] = [];
-        }
+  function createZonesAndFileSharePathsMap(
+    newZonesAndFileSharePathsMap: ZonesAndFileSharePathsMap,
+    item: FileSharePath
+  ) {
+    // Zones first
+    // If the zone doesn't exist in the map, create it
+    if (!newZonesAndFileSharePathsMap[`zone_${item.zone}`]) {
```
krokicki, line 73: It might be a good idea to have reusable utility functions which encode data into map keys, e.g.:
* zone name -> "zone_{zone name}"
* fsp name -> "fsp_{fsp name}"
These functions can then be reused when reading preferences and looking things up in this map. The benefit of making them functions is that if the logic for creating keys ever changes, it can be updated in a single place instead of having to search through the entire codebase. 
Also, you could call the function once and save the key into a variable, instead of recreating it every time it's needed.

src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
+      } as Zone;
+    }
+    // If/once zone exists, check if file share path exists in the zone
+    // And update zone accordingly
+    if (newZonesAndFileSharePathsMap[`zone_${item.zone}`]) {
+      const existingZone = newZonesAndFileSharePathsMap[
+        `zone_${item.zone}`
+      ] as Zone;
+      if (
+        !existingZone.fileSharePaths.some(
```
krokicki, line 86: I'm not sure if this check is necessary either. The backend should not return duplicate records for paths.

src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
+          (existingItem: FileSharePath) => existingItem.name === item.name
+        )
+      ) {
+        existingZone.fileSharePaths.push(item);
+        newZonesAndFileSharePathsMap[`zone_${item.zone}`] = existingZone;
```
krokicki, line 91: You don't need to put the zone back into the map. It should already be updated based on the previous line.

src/contexts/ZoneBrowserContext.tsx
```diff
@@ -34,77 +38,103 @@ export const ZoneBrowserContextProvider = ({
...
-        console.error('An unknown error occurred');
+        console.error('An unknown error occurred when fetching zones');
       }
     }
   }
+
+  // When app first loads, fetch file share paths
+  // and create a map of zones and file share paths
+  React.useEffect(() => {
+    if (Object.keys(zonesAndFileSharePathsMap).length === 0) {
```
krokicki, line 124: Not sure I understand the purpose of this `if`. I guess this only loads the data if the map is empty, which means only at the first page load? Wouldn't this prevent new data from being loaded when we navigate between pages or other refresh the view?

krokicki (commented):
Great start! Some of my thoughts below.

src/contexts/ZoneBrowserContext.tsx
allison-truhlar: Fixed in commit abb7774fb4c40b8d71c6474fa2ec0ff344d72913

src/contexts/ZoneBrowserContext.tsx
allison-truhlar: Refactored in commit d591c85b328767be147777109e57afd5ae593e46

src/contexts/ZoneBrowserContext.tsx
allison-truhlar: Fixed in commit d591c85b328767be147777109e57afd5ae593e46

src/contexts/ZoneBrowserContext.tsx
allison-truhlar: Added this utility function in commit d591c85b328767be147777109e57afd5ae593e46

src/contexts/ZoneBrowserContext.tsx
allison-truhlar: Fixed in commit d591c85b328767be147777109e57afd5ae593e46

allison-truhlar:
@krokicki Do the PreferencesContext look on the right track? See commit https://github.com/JaneliaSciComp/fileglancer/pull/43/commits/fafb0a8c4ae07c20f0d7a5392c59091233ad282f

krokicki:
> @krokicki Do the PreferencesContext look on the right track? See commit [fafb0a8](https://github.com/JaneliaSciComp/fileglancer/commit/fafb0a8c4ae07c20f0d7a5392c59091233ad282f)

Much improved! I can't build this PR right now though, there is some kind of build error.

allison-truhlar:
@krokicki The zone favorites should now be working on this branch. I decided to move forward with [saving the zone favorite data](https://github.com/JaneliaSciComp/fileglancer/blob/c6924d1ea130ba74c032ef6e2f96c08ede802a88/src/contexts/PreferencesContext.tsx#L160-L170) in the same format at the [`Zone` type ](https://github.com/JaneliaSciComp/fileglancer/blob/c6924d1ea130ba74c032ef6e2f96c08ede802a88/src/shared.types.ts#L23)used in the "uber" map, as opposed to the format we previously discussed: `{type: zone, name: fsp.name}`. The reason being I thought this would reduce the number of different states to keep track of. If we used the favorite-specific format, then I thought I would need to save these abbreviated data in a separate state to check whether a specific entry exists when updating the favorites, as well as separate state that stores the zones (or FSPs or folders) that gets passed to the components to render the favorites. Does this seem right to you? Is there any downside to saving the whole zone (or FSP) object to preferences, rather than just the abbreviated format?

krokicki:
Hi @allison-truhlar, 

The reason for the expanded format was to make it easier to reconstitute the data. If you only have the key then you have to parse it (i.e. split on underscores). It will work to use the keys, but it introduces fragility and certain limitations. For example, it prevents the use of underscores in the values moving forward.

By the way, I was able to build the branch, but I can't see the file share paths. They seem to load fine in the console, and the zones are displayed, but it seems like they're all empty:

Screenshot 2025-05-16 at 8 37 02 AM

Also, it's still saving the entire object as a preference. The idea was to save as little information as possible, to avoid having data in multiple places that can be desynced, and also for efficiency. 

Screenshot 2025-05-16 at 8 39 07 AM

Here, the preference should just be `{value: ["zone_Abbot", "zone_Beyene"]}`. The rest of the information can be reconstituted when loading the preferences.

krokicki (changes requested):
Not sure if it was the merge, but there are several things broken. I fixed one syntax issue, but there are a few in FavoritesBrowser that I can't resolve (it uses the type `FolderFavorite` which is not defined anywhere). 

These issues are being caught if you run `npm run eslint:check` but for some reason we're not running that in the CI. We should probably enable that and prettier and make a habit of using both before pushing code (or even add a precommit).

PR #44, opened 2025-05-14
Neuroglancer integration

krokicki:
I'm integrating [ome-zarr.js](https://www.npmjs.com/package/ome-zarr.js) and Neuroglancer. So far it identifies when you are in an OME-Zarr image by looking for a `.zattrs` file with a `multiscales` attribute. If if finds one, it shows a thumbnail of the image using ome-zarr.js and creates a button with a Neuroglancer link. This link loads data from the user server for now. 

This PR also contains some unrelated minor fixes:
* Use UNIX username instead of the Jupyter one for consistency. 
* Refactored out properties panel tab content into components (in preparation for adding OME-Zarr display, but I ended up putting it in the FileList)

I would appreciate a review, especially of the React code.

In order to integrate ome-zarr.js, I had to switch the `moduleResolution` from `node` to `bundler` in `tsconfig.json`. This worked, but currently there are a lot of warnings during `npm watch` that still need to be addressed. 

@StephanPreibisch @cgoina


src/contexts/FileBrowserContext.tsx
```diff
@@ -107,15 +116,85 @@ export const FileBrowserContextProvider = ({
...
+    try {
+      const fileText = await fetchFileAsText(path);
+      if (fileText === null) {
+        console.warn(`No text content fetched for path: ${path}`);
+        return null;
+      }
+      return JSON.parse(fileText);
+    } catch (error: unknown) {
+      if (error instanceof SyntaxError) {
+        console.error(`JSON parsing error for path ${path}: ${error.message}`);
```
krokicki, line 175: I think we need a better error handling scheme which does more than just log the errors.

fileglancer/handlers.py
```diff
@@ -66,10 +77,17 @@ def get(self):
...
+
 class FileShareHandler(APIHandler):
     """
     API handler for file access using the Filestore class
     """
+    def set_cors_headers(self):
+        self.set_header('Access-Control-Allow-Origin', '*')  # Or your specific domain
+        self.set_header('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, PATCH, OPTIONS')
+        self.set_header('Access-Control-Allow-Headers', 'Content-Type, X-XSRFToken')
+        self.set_header('Access-Control-Allow-Credentials', 'true')
```
krokicki, line 89: Vibe-coded CORS headers, need to be refined.

fileglancer/handlers.py
```diff
@@ -93,15 +111,21 @@ def _get_filestore(self, path_name):
         return filestore
 
 
-    @web.authenticated
```
krokicki, line 96: This is temporary, until we can get a file proxy in the central server.

fileglancer/handlers.py
```diff
@@ -24,6 +24,17 @@ def _get_mounted_filestore(fsp):
...
+def get_current_user(api):
+    """
+    Get the current user's username.
+    
+    Returns:
+        str: The username of the current user.
+    """
+    #return api.current_user.name
+    return os.getenv("USER") or os.getenv("USERNAME")
+
```
neomorphic, line 36: What about subclassing APIHandler, adding this as a class method and then falling back to self.current_user.name if the os.getenv() functions fail to return a value?

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "jsx": "react",
     "lib": ["DOM", "ES2018", "ES2020.Intl"],
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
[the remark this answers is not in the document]
neomorphic: This will be used by vite to build the frontend code. I will test once the code is merged

fileglancer/handlers.py
neomorphic: What was the auth preventing? Without this present, anyone can request that endpoint and get a response.

fileglancer/handlers.py
neomorphic: I am guessing we don't need to the  POST, PUT, DELETE, PATCH verbs. This is a read only end point, correct? Having the wildcard Origin might be necessary if, but if we are only using this to serve up content for our captive neuroglancer, then we can lock it down to that domain.

src/components/ui/FileBrowser/FileList.tsx
```diff
@@ -1,9 +1,13 @@
 import * as React from 'react';
 import { Typography } from '@material-tailwind/react';
+import * as zarr from "zarrita";
+import * as omezarr from "ome-zarr.js";
 
 import type { File, FileSharePathItem } from '../../../shared.types';
 import FileListCrumbs from './Crumbs';
 import FileRow from './FileRow';
+import { useZoneBrowserContext } from '../../../contexts/ZoneBrowserContext';
+import { useFileBrowserContext } from '../../../contexts/FileBrowserContext';
```
neomorphic, line 10: Not something for you to correct, but we should set up aliases in the bundler, so that we don't have to use relative paths. This should really be `~/contexts/FileBrowserContext`

src/contexts/FileBrowserContext.tsx
allison-truhlar: Added a [Clickup ticket]( https://app.clickup.com/t/86a8mm9au)

src/components/ui/FileBrowser/FileList.tsx
```diff
@@ -41,25 +45,72 @@ export default function FileList({
...
     return hideDotFiles
       ? files.filter(file => !file.name.startsWith('.'))
       : files;
   }, [files, hideDotFiles]);
 
-  const hasZattrs = files.some(file => file.name === '.zattrs');
+  const [hasMultiscales, setHasMultiscales] = React.useState(false);
+  const [thumbnailSrc, setThumbnailSrc] = React.useState<string | null>(null);
+  const [neuroglancerUrl, setNeuroglancerUrl] = React.useState<string | null>(null);
+  const neuroglancerUrlPattern = "https://neuroglancer-demo.appspot.com/#!{%22layers%22:[{%22source%22:%22zarr://{URL}%22,%22name%22:%22OME-NGFF%22}]}";
```
neomorphic, line 49: We are going to need to factor this out into something like an ENV variable. Future problem

src/components/ui/FileBrowser/FileList.tsx
```diff
@@ -41,25 +45,72 @@ export default function FileList({
...
+        if (zattrsContent && 'multiscales' in zattrsContent) {
+          setHasMultiscales(true);
+          const fileFetchPath = getFileFetchPath(currentNavigationPath.replace('?subpath=', '/'));
+          const imageUrl = `${window.location.origin}${fileFetchPath}`;
+          try {
+            const store = new zarr.FetchStore(imageUrl);
+            const src = await omezarr.renderThumbnail(store);
+            setThumbnailSrc(src);
+          } catch (error) {
+            console.error('Error rendering thumbnail', error);
```
neomorphic, line 67: Should we report back to the user on why the thumbnail is missing? We should probably add a stash for alerts, that can be rendered on any page.

src/components/ui/FileBrowser/FileList.tsx
```diff
@@ -41,25 +45,72 @@ export default function FileList({
...
+  }, [files, fetchFileAsJson, currentFileSharePath]);
 
   return (
     <div
       className={`px-2 transition-all duration-300 ${showPropertiesDrawer ? 'mr-[350px]' : ''}`}
     >
       <FileListCrumbs />
 
-      {hasZattrs && (
+      {hasMultiscales && (
```
neomorphic, line 85: The `variable && <JSX>` pattern can lead to '0' being rendered on the page. I have moved to using `variable ? <JSX> : null`

src/components/ui/FileBrowser/FileList.tsx
```diff
@@ -21,18 +26,83 @@ type FileListProps = {
...
+          }
+
+          const neuroglancerUrl = neuroglancerUrlPattern.replace("{URL}", imageUrl);
+          setNeuroglancerUrl(neuroglancerUrl);
+        }
+      }
+    };
+
+    checkZattrsForMultiscales();
+  }, [files, fetchFileAsJson, currentFileSharePath]);
```
allison-truhlar, line 77: Optional, but this `useEffect` could be moved to `FileBrowserContext.tsx`; then only the state values would need to be shared with the UI file. I've been trying to move most of the logic out of the UI files.

allison-truhlar (approved):
I left one comment regarding potentially moving the `useEffect` for the Zarr check in `FileList`. I'm happy to do the commit if you'd like to move forward with the change, or we can just leave it for now.

krokicki:
> I left one comment regarding potentially moving the `useEffect` for the Zarr check in `FileList`. I'm happy to do the commit if you'd like to move forward with the change, or we can just leave it for now.

Great idea, please go for it if you like!

src/contexts/FileBrowserContext.tsx
```diff
@@ -120,7 +120,7 @@ export const FileBrowserContextProvider = ({
     const url = getFileFetchPath(path);
 
     try {
-      const response = await sendGetRequest(url, cookies['_xsrf']);
+      const response = await sendFetchRequest(url, 'GET', cookies['_xsrf']);
```
neomorphic, line 123: We should add some unit testing. Not sure how that slips through the cracks

neomorphic (commented):
Added a few comments, mostly things to think about for the future

neomorphic (approved):

src/components/ui/FileBrowser/FileList.tsx
krokicki: Yes, the error handling throughout needs a revamp. Allison made a ticket for it.

src/components/ui/FileBrowser/FileList.tsx
krokicki: Fixed. Great tip, thanks!

src/contexts/FileBrowserContext.tsx
krokicki: Something went wrong with a merge, I think. Finding out what happened using `git bisect` is an exercise left to the reader.

fileglancer/handlers.py
krokicki: How delightfully object oriented! Fixed.

krokicki:
I'm going to go ahead and merge this so that we can continue development with other PRs.

PR #47, opened 2025-05-16
convert_build_to_vite

neomorphic:
This converts the whole build system for the UI to vite. From this point on the UI will be served by the backend extension at the /fg/ url. I have also modified the fileglancer icon on the launcher page to redirect the browser to the /fg/ url. All the hub code has been moved to a separate repo called [fileglancer-hub](https://github.com/JaneliaSciComp/fileglancer-hub). 

- **feat: Adds custom hub_launcher code.**
- **fix: removes hard coded filepaths from jupyterhub_config**
- **chore: Update tslint to use react-jsx**
- **fix: Corrects the default url.**
- **fix: Adds @web auth around the static files handler.**
- **fix: Sets the spawner command to jupyterhub-singleuser**
- **fix: prevents access to /user/<username>/lab**
- **fix: Removes unused imports of React**
- **chore: introduces a new tsconfig for vite.**
- **chore: Converts frontend build to vite.**
- **chore: Moved hub launcher code to a separate repo:**
- **chore: removes fileglancer/ui directory from VC**
- **chore: simplifies build of UI code.**
- **fix: restores build steps to get fileglancer Icon back in the lab.**
- **chore: removes tsc from watch src.**
- **chore: restore zarr packages after rebase error.**


krokicki:
Builds for me and looks great. The only issue I see is that the CI isn't working. 

The Copilot review is an interesting feature, but it looks like it mostly produced nonsense in this case. I'll be very curious if you think any of those comments are useful.

package.json
allison-truhlar, line 1: I got one warning when running `pixi run dev-watch`:
```
> vite build --watch

vite v6.3.5 building for production...

watching for file changes...

build started...
(node:56698) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///Users/truhlara/dev/fileglancer/postcss.config.js?t=1747663146093 is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /Users/truhlara/dev/fileglancer/package.json.
```
I tried adding `"type": "module"` to `package.json` - it does get rid of the warning and doesn't break the build/cause other errors. I looked in the diff for `package.json` and it doesn't look like you removed `"type": "module"` at any point, so maybe it's something to consider adding?

allison-truhlar (approved):
The build works for me - I especially appreciate the faster refresh for viewing style/layout changes!

PR #49, opened 2025-05-19
Fix sidebar scroll

allison-truhlar:
In this PR, I add a custom media query for short screens and then apply styles to get extra vertical space in the sidebar. 

It saves enough space to make the scroll in the favorites and zone browser still usable at the height at which the Starfish zone browser is no longer usable (around 290px).

Fileglancer at 284px screen height:
fileglancer zone browser at 284px

Starfish at 284px screen height:
starfish zone browser at 284px

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,5 +134,8 @@ dmypy.json
 pixi.lock
 package-lock.json
 
+# Zarr files
+*.zarr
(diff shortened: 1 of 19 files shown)


krokicki (approved):
Works great!

PR #50, opened 2025-05-19
chore: Adds src path alias to vite config.

neomorphic:
This adds the '@' alias to the build configuration, so that we no longer
have to figure out how far down the component tree we are when
importing another component. eg:

import { useZoneBrowserContext } from '../../../contexts/ZoneBrowserContext';

becomes

import { useZoneBrowserContext } from '@/contexts/ZoneBrowserContext';

@krokicki


allison-truhlar (approved):
Looks good and works for me - thanks for adding this. I added one commit where I changed a few more `../` paths to `@`.

PR #51, opened 2025-05-20
chore: Adds ui directory to dist artifacts.

neomorphic:
Without this, the ui directory is not packaged with the dist and doesn't
show up when installed in a production setting.


neomorphic:
@krokicki

PR #52, opened 2025-05-20
Proxied paths endpoints

cgoina:
@krokicki please take a look at the changes for the unit tests. I made some changes be able to control better what the API is server plugin is supposed to return. Also for the requests to central server I am using requests_mock. 

I still want to add unit tests for delete/create/update and for the response.raise_for_status() use case because I found some unexpected behavior while testing the unit tests.


fileglancer/proxiedpath.py
```diff
@@ -0,0 +1,96 @@
...
+            params = pp_updates
+        )
+
+    def _cache_user_proxied_paths(self, username: str) -> list[ProxiedPath] | None:
+        """Retrieve and cache user ."""
+        if self.central_url and username:
+            log.info(f"Cache proxied paths for user {username} from central server")
+            response = requests.get(f"{self.central_url}/proxied-path/{username}")
+            response.raise_for_status()
+            pps_json = response.json()["paths"]
```
cgoina, line 69: @krokicki I think the central API will require a change for this to return the proxied paths just like shared paths are returned.

conftest.py
```diff
@@ -1,8 +1 @@
-import pytest
-
 pytest_plugins = ("pytest_jupyter.jupyter_server", )
```
cgoina, line 1: I kept this conftest simple and added the proper config under fileglancer/tests

fileglancer/handlers.py
```diff
@@ -33,7 +34,7 @@ def get_current_user(self):
         Returns:
             str: The username of the current user.
         """
-        return os.getenv("USER", self.current_user.name)
+        return os.getenv("USER", self.current_user.username)
```
cgoina, line 37: name is actually user's name when it doesn't come from os.getenv("USER")

fileglancer/tests/test_localfilesharepaths.py
cgoina, line 1: renamed the unit test because the name reflects better what's happening inside

fileglancer/tests/test_mock_server.py
cgoina, line 1: This was renamed to test_filesharepaths.py - I don't know why git did pick up the rename

fileglancer/uimodels.py
```diff
@@ -0,0 +1,76 @@
+from typing import Optional, Generic, TypeVar
+from datetime import datetime, timedelta
+from pydantic import BaseModel, Field
+
+
+# TODO:
+# These are candidates to be extracted to a shared library
+# They have all been copied from fileglancer-central/fileglancer-central/model.py
+
+class FileSharePath(BaseModel):
```
cgoina, line 10: This used to be in paths.py but I moved it to a different file in case these will be all moved to a lib later

fileglancer/tests/test_proxiedpath.py
```diff
@@ -0,0 +1,155 @@
...
+        assert False, "Expected 404 error"
+    except Exception as e:
+        assert e.code == 404
+        rj = json.loads(e.response.body)
+        assert rj["error"] == "Proxied path not found"
+
+
+@patch("fileglancer.handlers.ProxiedPathHandler.get_current_user", return_value=TEST_INVALID_USER)
+async def test_get_user_proxied_path_when_central_responds_with_404(test_current_user, jp_fetch, requests_mock):
+    url_for_invalid_user = f"{TEST_CENTRAL_SERVER}/proxied-path/{TEST_INVALID_USER}"
```
krokicki, line 108: Nice to be able to test this!

fileglancer/proxiedpath.py
```diff
@@ -0,0 +1,98 @@
...
+class ProxiedPathManager:
+    """Manage the list of user shared data paths from the central server."""
+
+    def __init__(self, central_url: str):
+        self.central_url = central_url
+        self._cached_proxied_paths = {}
+    
+    def get_proxied_paths(self, username: str, sharing_key: Optional[str] = None) -> list[ProxiedPath]:
+        """Lookup a user shared path by its sharing key."""
+        cached_user_proxied_paths = self._cached_proxied_paths.get(username)
```
krokicki, line 22: I'm not sure we need to cache these proxied paths at all. The reason we cache the file share paths is because there was no way to get just one (maybe that should be added). But for proxied paths, we have full CRUD. Is there any other reason to cache these here?

fileglancer/paths.py
```diff
@@ -89,6 +57,7 @@ def __init__(self, central_url: str, dev_mode: bool, jupyter_root_dir: str):
                             FileSharePath(
                                 zone=zone,
                                 name=name,
+                                owner=zone.lower().replace(' ', '_'),
```
krokicki, line 60: Wasn't this field removed recently? File share paths don't have owners...

fileglancer/proxiedpath.py
krokicki: Yeah it currently returns a list, it's probably better to always return an object.

fileglancer/paths.py
cgoina: good catch - it should not be there

fileglancer/proxiedpath.py
cgoina: No reason at all - it was for consistency with the paths manager but I can remove it

fileglancer/handlers.py
```diff
@@ -332,7 +327,130 @@ def delete(self):
...
+            self.finish(json.dumps({"error": str(e)}))
+
+    @web.authenticated
+    def put(self):
+        """
+        Update a shared path for the current user.
+        """
+        username = self.get_current_user()
+        key = self.get_argument("key", None)
+        mount_path = self.get_argument("mount-path", None)
```
krokicki, line 372: It seems odd that the mount path is given as an argument called `mount-path` here, but it's in an object with key `mount_path` for the POST. Why not use an object here as well to be consistent? Also, the HTTP PUT verb is ideally supposed to use the body to update the object.

krokicki (approved):
Looks good, but I would change the PUT to use the body for consistency.

PR #54, opened 2025-05-21
Use request body to update proxied paths

cgoina:
Missed the comment to use the request body for ProxiedPath


krokicki (approved):
Thanks!

PR #57, opened 2025-05-23
Fix folder favs

allison-truhlar:


src/components/ui/Sidebar/Folder.tsx
```diff
@@ -59,8 +59,8 @@ export default function Folder({ folderFavorite, setOpenZones }: FolderProps) {
         'GET',
         cookies['_xsrf']
       );
-      const data = await response.json();
-      if (data.status === 200) {
+
+      if (response.status === 200) {
```
cgoina, line 63: @krokicki if this is all we need here - shouldn't there be a HEAD endpoint?

krokicki: Great idea. I'll add it to the issue tracker and we can come back to it later.

src/components/ui/FileBrowser/ContextMenu.tsx
```diff
@@ -61,7 +61,7 @@ export default function ContextMenu({
                 handleFavoriteChange(
                   {
                     type: 'folder',
-                    folderPath: `${currentFileSharePath.name}/${selectedFiles[0].path}`,
+                    folderPath: `${selectedFiles[0].path}`,
```
cgoina, line 64: can this just be selectedFiles[0].path - without using string interpolation

PR #58, opened 2025-05-23
Minor code improvements

krokicki:
Miscellaneous minor code improvements. 

@allison-truhlar @neomorphic @cgoina

### DIFF
--- a/src/contexts/PreferencesContext.tsx
+++ b/src/contexts/PreferencesContext.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import type { FileSharePath, Zone } from '@/shared.types';
 import { useCookiesContext } from '@/contexts/CookiesContext';
 import { useZoneBrowserContext } from './ZoneBrowserContext';
-import { getAPIPathRoot, sendFetchRequest, makeMapKey } from '@/utils';
+import { getAPIPathRoot, sendFetchRequest, makeMapKey, APIError } from '@/utils';
 
 export type FolderFavorite = {
   type: 'folder';
@@ -93,10 +93,13 @@ export const PreferencesProvider = ({
       ).then(response => response.json());
       return data?.value;
     } catch (error) {
-      console.log(
-        `Potential error fetching preferences, or preference with key ${key} is not set:`,
-        error
-      );
+      if (error instanceof APIError && error.responseCode === 404) {
+        console.log(`Preference '${key}' not found`);
+        return null;
+      }
+      else {
+        console.log(`Error fetching preference '${key}':`, error);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/contexts/PreferencesContext.tsx
```diff
@@ -93,10 +93,13 @@ export const PreferencesProvider = ({
       ).then(response => response.json());
       return data?.value;
     } catch (error) {
-      console.log(
-        `Potential error fetching preferences, or preference with key ${key} is not set:`,
-        error
-      );
+      if (error instanceof HTTPError && error.responseCode === 404) {
+        console.log(`Preference '${key}' not found`);
```
krokicki, line 97: This hides the stack traces which used to be thrown when there are no preferences.

src/contexts/PreferencesContext.tsx
```diff
@@ -175,10 +178,10 @@ export const PreferencesProvider = ({
 
     (async function () {
       const backendPrefs = await fetchPreferences('zone');
-      const zoneArray = backendPrefs.map((pref: ZonePreference) => {
+      const zoneArray = backendPrefs?.map((pref: ZonePreference) => {
```
krokicki, line 181: I ran into an error at some point with this, and this along with the `|| []` resolved it, so I put it in each block.

src/contexts/PreferencesContext.tsx
```diff
@@ -265,91 +268,79 @@ export const PreferencesProvider = ({
...
-      zonePreferenceMap
-    ) as Record<string, ZonePreference>;
-
     try {
+      const key = makeMapKey('zone', item.name);
+      const updatedZonePreferenceMap = updatePreferenceList(
+        key,
+        { type: 'zone', name: item.name },
+        zonePreferenceMap
+      ) as Record<string, ZonePreference>;
```
krokicki, line 277: Just moving all this inside the try block to improve error handling.

src/contexts/PreferencesContext.tsx
```diff
@@ -265,91 +268,79 @@ export const PreferencesProvider = ({
...
   }
 
   async function handleFavoriteChange(
     item: Zone | FileSharePath | FolderFavorite,
     type: string
   ) {
-    switch (type) {
-      case 'zone':
-        try {
+    try {
```
krokicki, line 328: Moved the try block to the outside so it only has to be done once instead of for each case.

src/contexts/PreferencesContext.tsx
```diff
@@ -93,10 +93,13 @@ export const PreferencesProvider = ({
...
-        `Potential error fetching preferences, or preference with key ${key} is not set:`,
-        error
-      );
+      if (error instanceof HTTPError && error.responseCode === 404) {
+        console.log(`Preference '${key}' not found`);
+        return null;
+      }
+      else {
+        console.log(`Error fetching preference '${key}':`, error);
+      }
```
cgoina, line 102: what do you return in this branch? does it make sense to move the return out of the if?

PR #60, opened 2025-05-26
adds currently logged in username

neomorphic:
- **chore: adds deprecation notice to setup_handlers function.**
- **feat: Adds username to profile menu.**

### DIFF
--- a/fileglancer/handlers.py
+++ b/fileglancer/handlers.py
@@ -541,6 +541,26 @@ class VersionHandler(BaseHandler):
         self.write(json.dumps({"version": version}))
         self.finish()
 
+class ProfileHandler(BaseHandler):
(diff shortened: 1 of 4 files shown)


fileglancer/handlers.py
```diff
@@ -595,6 +615,10 @@ def get(self, path):
 
 def setup_handlers(web_app):
     """
+    ! Deprectated: This function has been replaced by the `initialize_handlers` function.
+    ! Use `initialize_handlers` in app.py to set up the URL handlers for the Fileglancer
+    ! extension.
```
krokicki, line 620: Why deprecate instead of just delete?

krokicki (approved):
LGTM

PR #63, opened 2025-05-27
feat: Adds alerts via react-hot-toast.

neomorphic:
This is an example update to show how alerts can be added to the page,
without having to be embedded in the component where the causative
action has occurred. If this library is used, then we can change all
instances of the Alert component into calls to toast.

### DIFF
--- a/src/components/ui/Dialogs/SharingDialog.tsx
+++ b/src/components/ui/Dialogs/SharingDialog.tsx
@@ -1,16 +1,15 @@
 import React from 'react';
 import {
-  Alert,
   Button,
   Dialog,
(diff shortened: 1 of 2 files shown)


allison-truhlar (approved):
Thanks for looking into this, @neomorphic. This looks good to me - definitely an improvement over the user needing to manually close the alert.

allison-truhlar:
@neomorphic I left the alerts for add a new folder and rename folder, since I think these sometimes have errors that the user can correct easily in the input (e.g., trying to create a folder with the same name as another folder). Thoughts?

PR #66, opened 2025-05-29
Add loglevel library for logging

krokicki:
I added [loglevel](https://www.npmjs.com/package/loglevel) for better control of logging in the browser console. This will let us control levels in different environments (dev vs prod) and across components. 

For now, I just implemented logger usage in just omezarr-helper, as an example. 

Setting the default Vite environment to "development" has exposed a number of warnings from React. For example, there are a number of inline SVGs which cause issues like this:
```
Warning: Invalid DOM property `stroke-width`. Did you mean `strokeWidth`?
```
This could be addressed by something like https://www.npmjs.com/package/vite-plugin-svgr
There are also other warnings like this:
```
Warning: Each child in a list should have a unique "key" prop.
```
These should probably be addressed somehow before we merge this.

We also need a way to set the environment to "production" for the production deployment. 

@allison-truhlar @StephanPreibisch @cgoina


krokicki:
@allison-truhlar It was just a weird blip. I reran the job and it worked.

krokicki:
@cgoina Could you please take a look at why some of the new tests are failing here?

krokicki:
@cgoina Thanks but it looks like some tests are still failing.

cgoina:
@krokicki the UI tests are somewhat flaky - I don't understand why that one test kept failing even though it runs fine locally. I made some changes to the playwright settings and now it runs on the CI, but it's not 100% reliable.

PR #70, opened 2025-05-30
Fix build: sort entries to have dirs first

cgoina:
I noticed lately that the build fails. The reason is that os.listdir does not always return the files in the same order and lately it started to list the file in the unit test before the dir. This makes the Filestore yield function more deterministic by always returning the dirs firs.

### DIFF
--- a/fileglancer/filestore.py
+++ b/fileglancer/filestore.py
@@ -78,7 +78,7 @@ class Filestore:
         self.root_path = os.path.abspath(file_share_path.mount_path)
 
 
-    def _check_path_in_root(self, path: str) -> str:
+    def _check_path_in_root(self, path: Optional[str]) -> str:
         """
         Check if a path is within the root directory and return the full path.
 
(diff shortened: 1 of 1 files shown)


fileglancer/filestore.py
```diff
@@ -120,7 +120,8 @@ def get_file_info(self, path: Optional[str] = None) -> FileInfo:
         """
         full_path = self._check_path_in_root(path)
         stat_result = os.stat(full_path)
-        return FileInfo.from_stat(path, full_path,stat_result)
+        rel_path = os.path.relpath(full_path, self.root_path)
+        return FileInfo.from_stat(rel_path, full_path, stat_result)
```
cgoina, line 124: This change was mostly for type check

fileglancer/tests/test_filestore.py
```diff
@@ -54,7 +54,7 @@ def test_get_root_info(filestore, test_dir):
     file_info = filestore.get_file_info(None)
     assert file_info is not None
     assert file_info.name == os.path.basename(test_dir)
-    assert file_info.path is None
+    assert file_info.path == '.'
```
cgoina, line 57: @allison-truhlar do you know if we ever expect this to be null? If we do I can revert the change for get_root_info

fileglancer/filestore.py
krokicki: Looks like this converts None and empty string to a `.`, which is arguably more correct. I've added a comment to make that purpose clear in the code. Please let me know if I've misunderstood.

krokicki:
Thanks, @cgoina !

PR #71, opened 2025-05-30
Updated the github action to use pixi

cgoina:
Please take a look at these updates that allows us to use pixi for executing github action jobs.

I created this branch from the playwright-tests so please ignore the changes that are because the added ui-tests.


.github/workflows/build.yml
```diff
@@ -3,7 +3,7 @@ name: Build
 on:
   workflow_dispatch:
   push:
-    branches: main
+    branches: ga-with-pixi
```
cgoina, line 6: This is just temporary - I wanted to test this without creating a PR

.github/workflows/build.yml
```diff
@@ -98,7 +108,6 @@ jobs:
 
   integration-tests:
     name: Integration tests
```
cgoina, line 110: There's no need to have the dependency on the build here since it's installing everything it needs

package.json
```diff
@@ -89,6 +89,7 @@
         "globals": "^15.15.0",
         "prettier": "^3.5.3",
         "rimraf": "^5.0.1",
+        "rollup": "^4.41.1",
```
cgoina, line 92: @neomorphic without this dependency the build on github kept failing with some @rollup error

pyproject.toml
```diff
@@ -124,7 +124,9 @@ dev-install = { depends-on = ["pip-install", "node-install", "ext-enable"] }
 dev-watch = { cmd = "npm run watch" }
 dev-launch = { cmd = "jupyter lab --autoreload" }
 dev-launch-remote = { cmd = "jupyter lab --autoreload --ip=0.0.0.0 --no-browser" }
-pypi-build = { cmd = "python -m build", depends-on = ["node-install"] }
+pypi-build = { cmd = "pip install build && python -m build", depends-on = ["node-install"] }
```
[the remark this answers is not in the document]
krokicki: Per `RELEASE.md`, you should `pixi run pip install build twine hatch` before running `pypi-build`. 

We should probably just make a new pixi env called "release" which has these packages.

pyproject.toml
```diff
@@ -114,7 +114,7 @@ test = { features = ["test"], solve-group = "default" }
 
 [tool.pixi.tasks]
 ext-clean = "rm -rf fileglancer/labextension && rm -rf .pixi/envs/default/share/jupyter/labextensions/fileglancer"
-pip-install = { cmd = "pip install -e .", depends-on = ["ext-clean"] }
```
krokicki, line 117: This dependency was needed at some point to prevent weird things happening in the dev environment. Are you sure it's not needed anymore?

pyproject.toml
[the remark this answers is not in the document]
krokicki: Per `RELEASE.md`, you should run `pixi run pip install build twine hatch` before attempting to run `pypi-build`. We should probably just have another pixi environment for it. 

But I'm also not sure why `pypi-build` is necessary for CI actions at all?

krokicki (commented):
Looks promising. I'm just unsure about the changes made to current dev process. Does everything still work in dev?

pyproject.toml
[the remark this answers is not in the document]
cgoina: From what I see it is needed before uploading to pypi

allison-truhlar (approved):
Looks good to me! I just re-ran the integration test action that the PR failed on previously and it passed.

pyproject.toml
allison-truhlar: Also just came across the same error when trying to build a different branch on the dev server.

cgoina: @allison-truhlar can please add the dependency back if you're working on a branch where you found the problem.

krokicki: I already added it back on main: https://github.com/JaneliaSciComp/fileglancer/commit/49f6f973c6e71bdc22d2a6c70598c2c73557abe0

PR #73, opened 2025-06-03
Add favorite button to toolbar

krokicki:
I asked Claude Code to:
> Add a “favorite” button to the toolbar, which works the same way as the favorite button (star icon) in the zones list.

It seems to work.

@StephanPreibisch @neomorphic


neomorphic (commented):
Looks like it did the right thing. It might make things easier to maintain if the favorite "button" is encapsulated as a component that can be used in multiple locations within the app.

PR #74, opened 2025-06-06
Path display and handling

allison-truhlar:
Major changes in this PR:
- add standard path handling functions in `src/utils/pathHandling.ts`, and related edits to files to use these functions and remove ad-hoc path handling functions
- add unit tests for these functions in `src/__tests__/` These can be run via `npm run test`
- change `currentNavigationPath` to `currentFileOrFolder`. This state is now managed in `FileBrowserContext` as opposed to `ZoneBrowserContext`. 
  - **Note**: I haven't removed the changes I made to the `get` function in `FileShareHandler` to get the currentFileOrFolder data. This handler still needs to be refactored.
- replace `ZoneBrowserContext` with `ZonesAndFileSharePathMapContext`. This context only manages the state for the zone and file share path "uber" map
- use the new `getPreferredPathForDisplay` function to make sure all the paths displayed in the UI utilize the path preference set by the user
- set an initial file share path based on the first (alphabetically) file share path preference or, if none, on the first file share path within the first zone. This way, if the user navigates to `/browse` directly, they see the files inside this initial file share path rather than a blank screen.


fileglancer/handlers.py
allison-truhlar, line 1: **Note:** I haven't removed the changes I made to the get function in FileShareHandler to get the currentFileOrFolder data. This handler still needs to be refactored.

src/contexts/FileBrowserContext.tsx
```diff
@@ -35,87 +45,155 @@ export const FileBrowserContextProvider = ({
...
+      return;
     }
-  }, [makeDirArray, currentNavigationPath]);
 
-  React.useEffect(() => {
-    if (dirArray.length > 1) {
-      setCurrentDir(dirArray[dirArray.length - 1]);
-    } else {
-      setCurrentDir(dirArray[0]);
+    // Only set if currentFileSharePath is not set
```
allison-truhlar, line 151: Set an initial file share path based on the first (alphabetically) file share path preference or, if none, on the first file share path within the first zone. This way, if the user navigates to /browse directly, they see the files inside this initial file share path rather than a blank screen.

krokicki: I like the thought but in practice I think very few people will be interested in browsing the first zone, which is basically just one specific lab. 

What about using that initial screen for some kind of dashboard. We'd need to flesh this out, but some ideas:
* Link to home folder (based on user's group membership)
* Recently shared folders
* Recently viewed folders
* Most viewed folders

krokicki: I like the thought, but in practice I think very few people will be interested in the first zone, which is basically one specific lab.

What about using that space for a dashboard? We'd have to flesh this out, but some ideas:
* Home folder (based on group membership)
* Recently viewed folders
* Most viewed folders
* Recently shared folders

src/utils/pathHandling.ts
```diff
@@ -0,0 +1,88 @@
...
+    return url;
+  } else if (filePath) {
+    // Ensure the filePath is URL-encoded
+    filePath = encodeURIComponent(filePath);
+    url = joinPaths(url, filePath);
+  }
+  return url;
+}
+
+function getLastSegmentFromPath(itemPath: string): string {
```
krokicki, line 45: These types of functions would be really useful to briefly comment to describe the intent. I usually like to add an example of what it does given some usual input. That makes it much faster to read and understand what's happening later on.

src/utils/pathHandling.ts
```diff
@@ -0,0 +1,88 @@
...
+  return path.basename(itemPath);
+}
+
+function makePathSegmentArray(itemPath: string): string[] {
+  return itemPath.split(path.sep);
```
krokicki, line 50: Should we be using `path.sep` here? Won't this change based on the client OS, but our paths are always Linux paths?

krokicki (commented):
Looks great so far! I will do the server-side refactoring and let you know when it's ready to integrate.

neomorphic (approved):
Looks good to me

src/utils/pathHandling.ts
allison-truhlar: My assumption was that the client OS doesn't matter, since this is running in the browser. And since all the paths we're passing around in the UI code are in Linux style, until the point of conversion for display in the UI using the `getPreferredPathForDisplay` function, I figure it was safe to use `path.sep`.

src/utils/pathHandling.ts
allison-truhlar: I added comments in a new commit: 94efa55cff40e1989b88e32b53a851eecf48c994

src/contexts/FileBrowserContext.tsx
allison-truhlar: This is a good idea - I did notice this flaw when using the wiki FSPs. I added a placeholder Dashboard in this commit: [b9b607f](https://github.com/JaneliaSciComp/fileglancer/pull/74/commits/b9b607fe7ac2c7c3373b2183cf929cf3855c1bba)

src/utils/pathHandling.ts
krokicki: Thanks for checking! I'm a little uneasy depending on this since there are so many ways for it to mysteriously break in the future. I would recommend defining our own constant somewhere (`path_sep='/'`), that we can feel safe depending on.

src/utils/pathHandling.ts
krokicki: Nice!

src/utils/pathHandling.ts
allison-truhlar: Fixed this here: [06152fc](https://github.com/JaneliaSciComp/fileglancer/pull/74/commits/06152fc0763bcb8dfc747dc611bef76567834fcb)

src/contexts/ProxiedPathContext.tsx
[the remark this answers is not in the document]
krokicki: @neomorphic Can we resolve this or do you have any further concerns?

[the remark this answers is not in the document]
neomorphic: Please resolve. I just wanted to know about the decision that was made. I am happy with the code.

PR #76, opened 2025-06-09
fix: force y-scroll in file list when zarr preview is present

allison-truhlar:
@krokicki @neomorphic @cgoina 
This PR fixes the issue where, when the Zarr preview was present, the file browser list got cut off in short screens. Now, a y-scrollbar appears in these cases. I tested this in both Firefox and Chrome.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -17,3 +17,14 @@
     @apply w-3 h-3;
   }
 }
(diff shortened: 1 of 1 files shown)


krokicki:
Works nicely!

PR #78, opened 2025-06-11
Sharing UI

allison-truhlar:
Adds a "shared" page that lists the proxied file paths. For now, you can unshare a path and copy the path URL. 

After adding the URL params for file share path and file path, I will add the link to view a proxied path in "/browse".

### DIFF
--- a/src/components/Shared.tsx
+++ b/src/components/Shared.tsx
@@ -14,10 +14,7 @@ export default function Shared() {
           Shared Paths
         </Typography>
         <div className="rounded-lg shadow bg-background">
-          <div className="grid grid-cols-[0.5fr_1.5fr_2.5fr_1.5fr_0.5fr] gap-4 px-4 py-2 border-b border-surface">
-            <Typography variant="small" className="font-bold">
-              Unshare
-            </Typography>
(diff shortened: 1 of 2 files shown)


neomorphic:
We need to add some documentation in the README to make sure people know to create the .env file. If it is missing there is a cryptic error message when trying to share files and the shared page throws an error and whites out the site.

allison-truhlar:
> It might just be me, but it seems odd to have the unshare icon on one side of the table and the actions on the other. Could we have the dits and the unshare icon in the actions part of the table?

Fair point. Really, at this point, the "actions" icon could just be the unshare icon and a "copy" icon, since those are the only two actions.

src/utils/pathHandling.ts
```diff
@@ -1,5 +1,9 @@
 import path from 'path';
 import type { FileSharePath } from '@/shared.types';
+import type { ProxiedPath } from '@/contexts/ProxiedPathContext';
+
+const PATH_DELIMITER = '/';
+const PROXY_BASE_URL = import.meta.env.VITE_PROXY_BASE_URL;
```
neomorphic, line 6: In addition to adding documentation about adding this to an env file, we could throw a warning here if the import.meta.env... is undefined.

neomorphic (commented):
A few things to look at but overall looks good to me.

neomorphic (approved):
LGTM

PR #80, opened 2025-06-12
Fix shared context menu

allison-truhlar:
- Removes the "unshare" button on the Shared page
- Changes the "actions" icon on the Shared page to match the Browse page
- Switches to using the Material Tailwind `Menu` component on the Shared page
- Uses the Material Tailwind `Menu.Item` component, instead of styling that mimics it, for the items inside the context menu on the Browse page
- Adjusts the styling of the div that holds the context menu on the Browse page to use styles copied from the Material Tailwind `Menu` component


krokicki:
Looks great, I just updated some of the text and added a "copy path" option.

PR #81, opened 2025-06-13
set NODE_ENV during development

krokicki:
The NODE_ENV defaults to `production`, which sets the logging to warn. For the dev commands, I've set `NODE_ENV` to `development`. However, this doesn't seem to work, possibly because this environment is not being share with vite? @neomorphic do you have any ideas?


krokicki:
Nevermind, figured it out. I have to also set `NODE_ENV` when running npm.

PR #83, opened 2025-06-13
Error handling

allison-truhlar:
@neomorphic @krokicki  - no rush to look at this, but I would especially appreciate feedback on the possible error handling pattern before I continue implementing it for other functions/components.

This PR also adds an error boundary for catching unexpected rendering errors. I think this is pretty straightforward - I more or less followed vEMFlow but using the `react-error-boundary` library since I'm using the declarative version of `react-router`, which doesn't include the `useRouteError` hook.

**Possible error handling pattern:** After some back and forth with Claude for ideas, I added a possible error handling pattern for functions called by UI components, demonstrated by `handleNavigationClick`, defined inside `FileBrowserContext.tsx`. I've outlined it below - I welcome any feedback before I continue along this path.

- `handleNavigationClick` now returns an object of the format `{success: boolean; error?:string}`. 
- In the catch block of `handleNavigationClick`, a new utility function `createErrorResult` in `utils/errorHandling.ts` logs the caught error using the `loglevel` library and also return an error string to display in the UI via a the `react-hot-toast` library.
- Any functions called inside `handleNavigationClick` (e.g., `fetchAndFormatFilesForDisplay`) also return a success/error object, which is checked before continuing on with other logic. If it contains an error, `handleNavigationClick` returns early and passes along the object with the error to the UI.
- The UI does not use try/catch blocks - instead, it checks the value of the success/error object returned by `handleNavigationClick` and displays the error message if needed.

### DIFF
--- a/src/components/ui/FileBrowser/ContextMenu.tsx
+++ b/src/components/ui/FileBrowser/ContextMenu.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Menu, Typography } from '@material-tailwind/react';
+import toast from 'react-hot-toast';
 
 import type { FileOrFolder } from '@/shared.types';
 import { useFileBrowserContext } from '@/contexts/FileBrowserContext';
@@ -58,9 +59,9 @@ export default function ContextMenu({
         <Menu.Item>
           <Typography
             className="text-sm p-1 cursor-pointer text-secondary-light"
-            onClick={() => {
+            onClick={async () => {
               if (currentFileSharePath) {
-                handleFavoriteChange(
+                const result = await handleFavoriteChange(
                   {
                     type: 'folder',
                     folderPath: selectedFiles[0].path,
@@ -68,6 +69,9 @@ export default function ContextMenu({
                   },
                   'folder'
                 );
+                if (!result.success) {
+                  toast.error(`Failed to update favorites: ${result.error}`);
+                }
               }
               setShowContextMenu(false);
             }}
--- a/src/components/ui/Sidebar/FileSharePath.tsx
+++ b/src/components/ui/Sidebar/FileSharePath.tsx
@@ -71,9 +71,12 @@ export default function FileSharePathComponent({
           className="min-w-0 min-h-0"
           variant="ghost"
           isCircular
-          onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
+          onClick={async (e: React.MouseEvent<HTMLButtonElement>) => {
             e.stopPropagation();
-            handleFavoriteChange(fsp, 'fileSharePath');
+            const result = await handleFavoriteChange(fsp, 'fileSharePath');
+            if (!result.success) {
+              toast.error(`Failed to update favorites: ${result.error}`);
+            }
           }}
         >
           {isFavoritePath ? (
--- a/src/components/ui/Sidebar/Folder.tsx
+++ b/src/components/ui/Sidebar/Folder.tsx
@@ -134,9 +134,15 @@ export default function Folder({ folderFavorite, setOpenZones }: FolderProps) {
             className="min-w-0 min-h-0"
             variant="ghost"
             isCircular
-            onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
+            onClick={async (e: React.MouseEvent<HTMLButtonElement>) => {
               e.stopPropagation();
-              handleFavoriteChange(folderFavorite, 'folder');
+              const result = await handleFavoriteChange(
+                folderFavorite,
+                'folder'
+              );
+              if (!result.success) {
+                toast.error(`Failed to update favorites: ${result.error}`);
+              }
             }}
           >
             <StarFilled className="icon-small short:icon-xsmall mb-[2px]" />
--- a/src/components/ui/Sidebar/MissingFolderFavoriteDialog.tsx
+++ b/src/components/ui/Sidebar/MissingFolderFavoriteDialog.tsx
@@ -60,16 +60,18 @@ export default function MissingFolderFavoriteDialog({
               variant="outline"
               color="error"
               className="!rounded-md flex items-center gap-2"
-              onClick={() => {
-                try {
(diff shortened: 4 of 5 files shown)


neomorphic:
> @neomorphic @krokicki - no rush to look at this, but I would especially appreciate feedback on the possible error handling pattern before I continue implementing it for other functions/components.
> 
> This PR also adds an error boundary for catching unexpected rendering errors. I think this is pretty straightforward - I more or less followed vEMFlow but using the `react-error-boundary` library since I'm using the declarative version of `react-router`, which doesn't include the `useRouteError` hook.
> 

The error boundary code looks good to me.

> **Possible error handling pattern:** After some back and forth with Claude for ideas, I added a possible error handling pattern for functions called by UI components, demonstrated by `handleNavigationClick`, defined inside `FileBrowserContext.tsx`. I've outlined it below - I welcome any feedback before I continue along this path.
> 
> * `handleNavigationClick` now returns an object of the format `{success: boolean; error?:string}`.

It looks like handleFileBrowserNavigation is the function returning the above object.

> * In the catch block of `handleNavigationClick`, a new utility function `createErrorResult` in `utils/errorHandling.ts` logs the caught error using the `loglevel` library and also return an error string to display in the UI via a the `react-hot-toast` library.

Maybe we can have the error handling code post an error message to our server logs, so that we can see the errors and they aren't lost inside the users console. navigator.sendBeacon() can do this if we want to. With the addition of the createErrorResult function, we have to remember to call createErrorResult in our error blocks, maybe there is a way to wrap the code we want to prevent from throwing with a function that handles both the happy path and the error conditions, so that we don't have to remember what to put in a successful return or a failed return.  I got this from a quick chat with GPT4

```
async function handleWithError<T>(fn: () => Promise<T & { success: boolean }>): Promise<T | { success: false; error: any }> {
  try {
    const result = await fn();
    return result;
  } catch (error: unknown) {
    return createErrorResult('Navigation', error);
  }
}

return await handleWithError(async () => {
  const fetchFilesResult = await fetchAndFormatFilesForDisplay({
    fspName: fetchPathFsp as string,
    ...(path && { path })
  });

  if (!fetchFilesResult.success) {
    return fetchFilesResult;
  }

  if (!currentFileOrFolder || currentFileOrFolder.path !== path) {
    const updateResult = await updateCurrentFileOrFolder({
      fspName: fetchPathFsp as string,
      ...(path && { path })
    });
    if (!updateResult.success) {
      return updateResult;
    }
  }

  return { success: true };
});

```
 

> * Any functions called inside `handleNavigationClick` (e.g., `fetchAndFormatFilesForDisplay`) also return a success/error object, which is checked before continuing on with other logic. If it contains an error, `handleNavigationClick` returns early and passes along the object with the error to the UI.
> * The UI does not use try/catch blocks - instead, it checks the value of the success/error object returned by `handleNavigationClick` and displays the error message if needed.

I am fine with no try/catch blocks in the UI code, provided everything that is doing data fetching/manipulation is in a hook.

src/utils/errorHandling.ts
```diff
@@ -16,4 +20,16 @@ function createErrorResult(operation: string, error: unknown) {
   };
 }
 
-export { createErrorResult };
+async function tryCatchWrapper<T>(
```
allison-truhlar, line 23: @neomorphic Is this along the lines of what you were suggesting? I then used this in all the functions in `FileBrowserContext.tsx` that previously had success messages manually implemented. The repetition of it did make me wonder if my `handleFileBrowserNavigation` function is too complex, the way it requires three of these `tryCatchWrappers` - or is this typical?

src/utils/errorHandling.ts
```diff
@@ -6,6 +6,10 @@ function formatError(error: unknown): string {
 
 function logError(operation: string, error: unknown): void {
   log.error(`${operation} failed: ${formatError(error)}`);
+  //   navigator.sendBeacon('/log', JSON.stringify({
```
allison-truhlar, line 9: @neomorphic This is just pseudocode, but is this what you were suggesting? If so, am I correct in assuming it would require adding a new endpoint to the API, to handle these post requests?

src/hooks/useDeleteDialog.ts
```diff
@@ -16,30 +15,36 @@ export default function useDeleteDialog() {
...
+      if (!navResult.success) {
+        return {
+          success: false,
+          error: `Delete succeeded but navigation failed: ${navResult.error}`
+        };
+      }
+
+      return {
+        success: true,
+        targetPath: targetItem.path
```
krokicki, line 45: This defines a new result type with a `targetPath` property, which is inconsistent with other Results. We could put attributes in the data field to stay consistent with a single result schema. 

More generally, and to avoid situations like this in the future, rather than re-defining this object type every time it's needed, we could have finer-grained types and a factory to create instances.

For example:
```
export type Result<T> = Success<T> | Failure;

export interface Success<T> {
  success: true;
  data: T;
}

export interface Failure {
  success: false;
  error: string;
}

```

Then you can have a factory to help create them:
```
class ResultFactory {
  static ok<T>(data: T): Result<T> {
    return { success: true, data };
  }

  static fail(error: string): Failure {
    return { success: false, error };
  }
}
```

Then the code here becomes easier and more predictable:
```
      if (!navResult.success) {
        return ResultFactory.fail(`Delete succeeded but navigation failed: ${navResult.error}`)
      }
      return ResultFactory.ok({targetPath: targetItem.path})   
```

I would also add convenience methods for checking for success, so you don't have to access the success property directly. 

At this point though, I wondered if there are existing libraries that already implement these patterns and it turns out that there are: https://github.com/supermacro/neverthrow

@neomorphic what do you think?

allison-truhlar: @krokicki Thanks for looking into this - it does seem that what I'm doing is a less robust implementation of this library. I don't want to reinvent the wheel, and I also agree that my version is getting a little hard to follow.

@neomorphic Do you agree this library looks like it would be worth switching over to?

krokicki:
@allison-truhlar I noticed a new bug with this branch: when I click on Abbott->primary, that folder is empty but the loader icon remains spinning.

PR #85, opened 2025-06-17
fix: use URL constructor to ensure // after protocol

allison-truhlar:
@krokicki @neomorphic 
The `joinPaths` method was replacing the `//` with `/'. The `path.join` method in NodeJS first joins the path segments and then normalizes them - any double slashes are replaced with a single slash. Using the URL constructor maintains the double slash, plus these are URLs we're forming so it's probably better practice.


krokicki:
LGTM

neomorphic (approved):
LGTM. With tests too!! 🎊

PR #86, opened 2025-06-23
Feat: add url params to browse routes

allison-truhlar:
This PR replaces manual navigation logic, including setting file share path and file path context, with URL param-based navigation. This allows the URL to store the state of the file browser.

### File Browser Navigation Updates:
* `src/App.tsx`: Adds new routes to support navigation based on `fspName` and `filePath` URL parameters.
* `src/contexts/FileBrowserContext.tsx`: Removed manual navigation methods (`handleFileBrowserNavigation`, `updateCurrentFileOrFolder`) and replaced them with a `fetchFiles` method for loading files based on `fspName` and `filePath`. Updated the context provider to accept `fspName` and `filePath` as props.

### UI Component Navigation Improvements:
* Use React Router `Link` components for navigation. 
* Use the new `makeBrowseLink` function to construct navigation links with URL fspName and filePath params


src/utils/pathHandling.ts
```diff
@@ -180,10 +180,36 @@ function getPreferredPathForDisplay(
...
+  fspName: string | undefined,
+  filePath?: string
+): string {
+  if (!fspName) {
+    log.warn('FSP name is required to create a browse link.');
+    return '/browse';
+  }
+  return filePath
+    ? `/browse/${encodeURIComponent(fspName)}/${encodeURIComponent(filePath)}`
+    : `/browse/${encodeURIComponent(fspName)}`;
```
neomorphic, line 212: The encodeURIComponent function is messing up the urls. I am seeing this:
`http://localhost:8888/fg/browse/local/fileglancer%2Ftests%2Ftest_filesharepaths.py`
instead of `http://localhost:8888/fg/browse/local/fileglancer/tests/test_filesharepaths.py`

switching to encodeURI would preserve the '/'

src/hooks/useNewFolderDialog.ts
```diff
@@ -58,10 +57,7 @@ export default function useNewFolderDialog() {
       );
       try {
         await addNewFolder();
-        await handleFileBrowserNavigation({
-          fspName: currentFileSharePath.name,
-          path: currentFileOrFolder.path
-        });
+        await fetchFiles(currentFileSharePath.name, filePath);
```
neomorphic, line 60: For whatever reason, when I add a new folder, I see the toast to tell me that the folder was added, but I am not seeing it in the list, until I refresh the page.

src/hooks/useDeleteDialog.ts
```diff
@@ -11,26 +10,19 @@ import { useFileBrowserContext } from '@/contexts/FileBrowserContext';
...
-    );
+    const fetchPath = getFileBrowsePath(fspName, targetItem.path);
 
     try {
       await sendFetchRequest(fetchPath, 'DELETE', cookies['_xsrf']);
-      await handleFileBrowserNavigation({
-        fspName: currentFileSharePath.name,
-        path: removeLastSegmentFromPath(targetItem.path)
-      });
+      await fetchFiles(fspName, removeLastSegmentFromPath(targetItem.path));
```
neomorphic, line 25: When I delete a folder I also get the toast alert to say it has been deleted, but I am not seeing the updated file list with the folder missing.

src/components/ui/FileBrowser/Toolbar.tsx
```diff
@@ -54,44 +56,43 @@ export default function Toolbar({
...
     return folderKey in folderPreferenceMap;
-  }, [
-    currentFileSharePath,
-    currentFileOrFolder,
-    folderPreferenceMap,
-    fileSharePathPreferenceMap
-  ]);
+  }, [filePath, fspName, folderPreferenceMap, fileSharePathPreferenceMap]);
+
+  const link = makeBrowseLink(fspName, filePath);
```
neomorphic, line 70: It doesn't look like this link is used in the toolbar. Was it intended for something that hasn't been added yet?

src/components/ui/FileBrowser/Toolbar.tsx
```diff
@@ -54,44 +56,43 @@ export default function Toolbar({
...
 
   const handleFavoriteClick = React.useCallback(async () => {
     if (!currentFileSharePath || !currentFileOrFolder) {
       return;
     }
-    console.log('path:', currentFileOrFolder.path);
-    if (currentFileOrFolder.path === '.') {
+    console.log('path:', filePath);
```
neomorphic, line 76: remove?

neomorphic (commented):
Almost there, but I noticed a few bugs, which I have commented on in the code.

I saw some other bugs that I wasn't sure where to comment, so here they are:

Reloading a page when a file has been selected shows an empty filebrowser. Maybe clicking on a file should not add to the path, only folders. If clicking on a file is going to open the properties for that file, then it could add to the url, but the to code would have to attempt to load the folder information from a file path, fail, truncate the path and try again.
Screenshot 2025-06-23 at 5 53 18 PM

 Loading a file path that doesn't exist returns a blank file browser, it should probably return a 404
 
Screenshot 2025-06-23 at 17 59 07

src/utils/pathHandling.ts
allison-truhlar: Fixed in https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/a109b046d5f922a514b4f3d9015e8e838ca0db59

src/hooks/useDeleteDialog.ts
allison-truhlar: Fixed in https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/a422d3e8a13d0d637cee751b3aa400c7b06bcbd4

src/components/ui/FileBrowser/Toolbar.tsx
allison-truhlar: That was left behind from my first attempt at updating the "refresh" button in the toolbar. I deleted it: https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/6e5ce126d9933ae630b6a37367390185820ad9e0

allison-truhlar:
> Reloading a page when a file has been selected shows an empty filebrowser. Maybe clicking on a file should not add to the path, only folders. If clicking on a file is going to open the properties for that file, then it could add to the url, but the to code would have to attempt to load the folder information from a file path, fail, truncate the path and try again.

For now, I've made these changes:

- Added logic to `FileBrowserContext` to check if the file path URL param fetches a file - if it does, it then removes the last segment of the path and attempts the fetch again. If this results in a folder, then the contents of the folder are displayed (which would include the file that the user was initially attempting to view). 
- Files in the file browser are now text, instead of links, to make it harder for someone to generate a URL that would include a path to a file instead of a folder

I haven't added logic to make a file path URL that leads to a file result in open the properties panel with information for that file. This could be added in a new PR.

allison-truhlar:
> Loading a file path that doesn't exist returns a blank file browser, it should probably return a 404

I made changes so that if the result of fetching a file path is undefined, then the error boundary is shown: https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/b1a4995d95d62af756cde70316946cce54fd5641#diff-d13b6a5f5da5b1f8b9864ec8f29910b4e1a2c20748ffcf31d45c62c775f09054
To do this, I cherry-picked some commits from the `error-handling` branch: https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/fdbdb2081a5627e0e5bbf72a22cca104a8998857; https://github.com/JaneliaSciComp/fileglancer/pull/86/commits/62fc53cf791404fa100a9dc35785e959c732cded

neomorphic (approved):
LGTM

krokicki (approved):
LGTM

PR #90, opened 2025-06-26
feat: enhance error fallback UI

allison-truhlar:
- Include a link for submitting bug reports
- Replace error icon with one using Fileglancer colors


neomorphic (approved):
The icon looks much nicer with some color.

PR #93, opened 2025-07-07
Alternative error handling

allison-truhlar:
@krokicki Again, no rush to look at this, but I would appreciate your thoughts before I continue refactoring the code base to follow this implementation of the error handling.

**Note:** This PR is a draft/work-in-progress. I haven't refactored every function to follow this new pattern yet.
This PR adds what I think is a more straightforward implementation of error handling than that proposed in #83. Informed by this [blog post](https://www.haydenbleasel.com/blog/on-javascript-errors).

### Main changes:

- Adds `handleFailure` and `parseError` utility functions. `handleFailure` receives either response objects or errors, logs them, and then uses `parseError` to return a string description of the error, which is used to show a toast notification. 
- - Adds tests of `handleFailure` and `parseError`
- Following the **Just use try/catch** section of the blog post, removes try/catch blocks from lower level functions, allowing errors to bubble up to the UI level. This is where `handleFailure` is called.
- Refactors lower level functions making API calls to return a response, which can be handled at the UI level code, like the errors, by `handleFailure` if `!response.ok`.

### DIFF
--- a/src/__tests__/errorHandling.test.ts
+++ b/src/__tests__/errorHandling.test.ts
@@ -1,0 +1,125 @@
+import { describe, test, expect, afterEach, beforeEach, vi } from 'vitest';
+import { parseError, handleFailure } from '@/utils/errorHandling';
+
+// Mock loglevel
+vi.mock('loglevel', () => ({
+  default: {
+    error: vi.fn()
+  }
+}));
+
+// Mock the react-hot-toast import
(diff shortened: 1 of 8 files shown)


krokicki (commented):
Glad you are investigating this and trying different things. I'm not totally convinced by that blog post, but I'm not a Javascript/Typescript expert. Let's discuss further with the team before moving forward.

PR #94, opened 2025-07-07
Add GitHub link and update icon imports

allison-truhlar:
This PR adds a link to the GitHub repo in the navbar. It also updates the codebase to replace `@heroicons/react` icons with `react-icons` equivalents, removing `@herocions/react` as a dependency and thereby slightly reducing the final size of the app.


krokicki (approved):
LGTM

PR #95, opened 2025-07-08
Bug fix - error when Zarr thumbnail can't load

allison-truhlar:
This PR adds a try/catch block in getOmeZarrMetadata, specifically around the `omezarr.renderThumbnail` method. If this method throws an error, then it's caught in the catch block, and thumbnail is set to `null`. Then, `ZarrPreview.tsx`, if metadata exists but thumbnail does not, a placeholder is rendered in the UI stating that the thumbnail could not be loaded, and the rest of the metadata is shown as normal.

You can test this by changing the `maxThumbnailSize` in line 390 of `omezarr-helper.ts` to equal 100. This should print a warning in the console that the "Lowest resolution is too large for Thumbnail. Limit is 100*100", but otherwise you should be able to share the image and then view it in Neuroglancer.

@krokicki @neomorphic


src/components/ui/FileBrowser/ZarrPreview.tsx
```diff
@@ -66,19 +66,25 @@ export default function ZarrPreview({
...
                 src={thumbnailSrc}
                 alt="Thumbnail"
                 className="max-h-72 max-w-max rounded-md"
               />
+            ) : !loadingThumbnail && metadata && !thumbnailSrc ? (
+              <div className="h-[225px] w-[225px] flex items-center bg-surface-light rounded-sm p-2">
+                <Typography className="text-foreground">
+                  Unable to load OME-Zarr image thumbnail.
+                </Typography>
+              </div>
```
cgoina, line 87: I think works but wouldn't it be better to have a "standard ZARR or N5" thumbnail for the case when the thumbnail cannot be created.

cgoina: that's correct

krokicki: That's the way Zarrcade works, when there is no thumbnail it just displays this image: https://github.com/JaneliaSciComp/zarrcade/blob/main/static/zarr.jpg

But for Zarrcade the user has to manually generate thumbnails if they want them. In the Fileglancer case, it should always be able to generate some sort of thumbnail from a valid OME-Zarr. When it can't it means that there is something kind of wrong with the OME-Zarr, for example there are not sufficiently small multiscale levels. I think putting that kind of error in the UI is probably a good idea. 

BioFileFinder uses a "broken image" placeholder, which might be the best of both worlds:
Screenshot 2025-07-08 at 4 42 00 PM
(although the particular image they chose is a little hard to parse visually)

allison-truhlar: Okay, thanks for the clarification! I can add a placeholder image and also display the specific error in the UI.

PR #98, opened 2025-07-11
Add "no permission" message in UI

allison-truhlar:
ClickUp id: #86aa3qxr0

@krokicki @neomorphic 
This PR checks for a 403 status and displays a permission error in this case, rather than a "no files to display" error.

I thought this would be straightforward fix, but I ran into a couple of issues:

First, I found that the `GET` method on the `FileMetadataHandler` was [not raising](https://github.com/JaneliaSciComp/fileglancer/blob/5fd82461d4885fd068ae0d9a9982d4ee308c1b5f/fileglancer/handlers.py#L197) the `FileNotFound` or `PermissionError` when I would navigate to a folder I did not have permission to view (e.g., [Eric's home folder](http://localhost:8888/fg/browse/groups_scicompsoft_home/trautmane)). The error was occurring in the initial call to `yield_file_infos` within the `GET` method, which was just [returning ](https://github.com/JaneliaSciComp/fileglancer/blob/5fd82461d4885fd068ae0d9a9982d4ee308c1b5f/fileglancer/filestore.py#L162)in the case of a `FileNotFound` or `PermissionError`, instead of raising them. Then I changed the way the file info was written to JSON, to prevent a JSON error in the case where the initial call to `yield_file_infos` failed, which used to leave an un-closed list for files (e.g., `"files": [`).

The second issue was that the `HTTPError` thrown by the `sendFetchRequest` method on the frontend was not being caught in the try/catch blocks of either [`fetchFileOrFolderInfo`](https://github.com/JaneliaSciComp/fileglancer/blob/5fd82461d4885fd068ae0d9a9982d4ee308c1b5f/src/contexts/FileBrowserContext.tsx#L69C23-L85C14) or [`fetchAndSetFiles`](https://github.com/JaneliaSciComp/fileglancer/blob/5fd82461d4885fd068ae0d9a9982d4ee308c1b5f/src/contexts/FileBrowserContext.tsx#L98) inside `FileBrowserContext.tsx`. I have a [commit](4da588664e11f52f9d650888e85fb9f31169c445) you can checkout to see this. This is the [commit](c967546b7f95f2df9b734b2fc833d91a01c3c508) with the workaround I ended up using.


fileglancer/filestore.py
```diff
@@ -158,9 +159,13 @@ def yield_file_infos(self, path: Optional[str] = None) -> Generator[FileInfo, No
                 try:
                     yield self._get_file_info_from_path(entry_path)
                 except (FileNotFoundError, PermissionError):
+                    log.error(f"Error accessing entry: {entry_path}")
                     continue
         except (FileNotFoundError, PermissionError):
-            return
+            if PermissionError:
+                raise PermissionError(f"Permission denied for path: {full_path}")
+            elif FileNotFoundError:
```
krokicki, line 167: Looking at this again, I don't think it will work. The `if` isn't testing anything here, it's always true because PermissionError is truthy. The correct way to write this would be to have two `except` statements, one for each type. But I don't think the try/except is necessary at all, because the code calling this is only looking to catch these types, so if we remove the try/except they will be thrown. I will fix this in another PR I'm working on, but please let me know if I missed something.

PR #99, opened 2025-07-14
Ticket workflow MVP

allison-truhlar:
ClickUp id: #86a84rg4u

This PR adds an MVP for the "create a ticket in JIRA" workflow. It includes the following:

Frontend:
- Add ConvertFile dialog that can be accessed from a button in the properties panel or the context menu. This dialog only handles one file for now, and only provides the option for the user to type in a destination path for the converted file
- Add TicketContext. Includes state for the ticket associated with the current "propertiesTarget" (defined in FileBrowserContext) and all tickets for the user. Includes methods to create a ticket, fetch one ticket's details based on fsp_name and path (assumes only one ticket per file), and fetch all tickets.
- Add a TicketDetails component to the PropertiesDrawer that displays the ticket details, if applicable, for a given propertiesTarget (file).
- Add the "Jobs" page, which has a table listing all jobs for the user

Backend (all analogous to Proxied Paths):
- Add a TicketManager to manage the list of tickets from fileglancer_central
- Modify the GET methods to allow fetching tickets by username, and optionally fsp_name and path

[Related changes in fileglancer_central](https://github.com/JaneliaSciComp/fileglancer-central/pull/20)

@neomorphic

### DIFF
--- a/src/components/Jobs.tsx
+++ b/src/components/Jobs.tsx
@@ -1,7 +1,7 @@
 import { Typography } from '@material-tailwind/react';
+
+import TicketRow from '@/components/ui/JobsPage/TicketRow';
 import { useTicketContext } from '@/contexts/TicketsContext';
-import { Link } from 'react-router';
-import { formatDateString } from '@/utils';
 
 export default function Jobs() {
   const { allTickets } = useTicketContext();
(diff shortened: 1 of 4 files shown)


krokicki (commented):
Fantastic work, I tried it out and it worked perfectly. I love the timeline view. 

Some minor enhancements we could make:
1) Let's use the user's chosen path formatting on the jobs page.
2) Inside the ticket, use the Linux path ("please convert /nrs/...") since this is information for people in SciCompSoft. 
3) In the ticket we should include the username of the requester.
4) This should probably be a separate ticket: when you click on the file link on the jobs page, it would be nice if it also opened the properties tab and navigated to the convert tab.

PR #104, opened 2025-07-16
Add unit test for search filter in sidebar component

allison-truhlar:
ClickUp id: 86a9dr3g1

This PR: 

- adds the dev dependencies react-testing-library and Mock Service Work to allow for unit tests of the components. 
- updates the Vite config to use a setup file, `src/__tests__/setup.ts`, which sets up the MSW server and allows you to use the jest-dom matchers used in react-testing-library with Vitest. 
- `src/__tests__/test-utils.tsx` sets up the `customRender` function, which wraps the component that's being rendered in a `<MemoryRouter>` (a `<BrowserRouter>` component for use in testing) and all the context providers that the app uses.
- Inside the `src/__tests__/mocks` directory, there are the files that setup the MSW server to intercept fetch requests and return mock data.
- Adds a unit test for the `<Sidebar>` component that confirms all the fake zones are rendered in the sidebar initially, and then only the matching zone is rendered after a simulated user event of typing in the search box.


neomorphic (approved):
LGTM. Nice use of msw. We will surely use that in the future for other tasks.

PR #106, opened 2025-07-17
fix: responsive formatting for error boundary

allison-truhlar:
ClickUp id: 86aa95dt5

Adds a y scroll bar to the error boundary. Also adds a minimum height to the navbar so it doesn't get covered by the banner/error message at very short screen sizes.


allison-truhlar:
> I think we should use this very sparingly, only in places where something goes so wrong that we can't recover. "Folder not found" is probably not one of those cases, and could be treated more like the permission messages.

Agreed - the error boundary showing up for an FSP you don't have access to was introduced by an Error being thrown in lines 130-131 in this commit: https://github.com/JaneliaSciComp/fileglancer/pull/108/commits/78600266837694c613a557c8248ad6c699983906, but it shouldn't be in the fetch files functions at all so I removed it.

This issue also highlights again for me I'm not sure when functions should be throwing Errors, so it would be helpful to have the discussion about error handling soon.

PR #107, opened 2025-07-17
Feat: add copy path to bread crumbs, toolbar and context menu

allison-truhlar:
ClickUp id: 86aa4k5dn

This PR:
- Adds a copy path icon to the end of the file browser bread crumbs, which appears when you hover over any part of the bread crumbs. In addition, the bread crumbs now display the file share path `mount_path` as the first segment of the file path, and also uses a separator that is appropriate to the path preference of the user.
- Adds a copy path button to the toolbar. 
- Adds "copy path" to the context menu.
- Refactors where the `copyToClipboard` function lives, moving it from a hook to a utility file. This is because we are no longer using the `<Alert>` component to show success/error messages, so I don't need to store state with the alert message. Without any associated state, the function can just be a utility function.

@krokicki @neomorphic

### DIFF
--- a/src/components/Browse.tsx
+++ b/src/components/Browse.tsx
@@ -58,7 +58,6 @@ export default function Browse() {
   return (
     <div className="flex-1 overflow-auto flex flex-col h-full">
       <Toolbar
-        selectedFiles={selectedFiles}
         hideDotFiles={hideDotFiles}
         setHideDotFiles={setHideDotFiles}
         showPropertiesDrawer={showPropertiesDrawer}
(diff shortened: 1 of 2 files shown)


krokicki (approved):
Awesome, very useful!

One very minor issue I noticed is when copying Mac paths, it only puts a single slash after the protocol:

```
smb:/nrs.hhmi.org/opendata/ome-zarr-examples/fly-efish/NP08_R2_20231017/NP08_R2_2_1_SS83741_Gpb5_546_ITP_647_1x_LOL.zarr/0/0
```

PR #110, opened 2025-07-18
fix: add link to go back to browse from error pg

allison-truhlar:
ClickUp id: 86aaau7pb

This PR fixes the problem of getting stuck on the error page by adding a link to `/browse`
@krokicki @neomorphic

### DIFF
--- a/src/components/ErrorFallback.tsx
+++ b/src/components/ErrorFallback.tsx
@@ -1,4 +1,5 @@
 import logger from '@/logger';
+import { Link } from 'react-router';
 import { Typography } from '@material-tailwind/react';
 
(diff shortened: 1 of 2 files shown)


krokicki (approved):
LGTM, but how can we test this now? Is there a way to trigger the error page?

allison-truhlar:
@krokicki If you add `throw new Error("Testing error boundary")` to any page component before the return statement, it should cause the error page to appear when you navigate to that page.
E.g., `Jobs.tsx`:
```
export default function Jobs() {
  throw new Error('Testing error boundary');
...existing code...
```

PR #111, opened 2025-07-18
Fix: y-scrolling on all pages

allison-truhlar:
Click up id: 86aaau0x6

This PR:
- Adds a layout for all pages other than Browse that allows y-scrolling when there is overflow. 
- Fixes y-scrolling the Browse page for the case of short screens and no file list (which already can scroll vertically), only the dashboard

@krokicki @neomorphic


krokicki (approved):
LGTM

PR #114, opened 2025-07-21
Fix miscellaneous bugs

allison-truhlar:
ClickUp id: 86aad2wx3

Fixes the following bugs/style errors:

- Adds margins to the preferences page. Achieved through updating the `OtherPagesLayout` so all child page components are flex columns with a padding of 6rem and the same maximum width.
- Removes errors in the console when you first load the page:
"No current file share path or file/folder selected Component Stack"
"Uncaught (in promise) Error: Ticket not found"
Achieved through changes to `TicketContext`, specifically changing error handling in `fetchAllTickets` and preventing the `useEffect` with `fetchTicket` from running if no file share path or file/folder are selected.
- Adds logic to `convertToMacPath` in the utilities to ensure that Mac paths already correctly formatted with a double slash after `smb:` don't get a third slash added.

### DIFF
--- a/src/utils/pathHandling.ts
+++ b/src/utils/pathHandling.ts
@@ -141,7 +141,9 @@ function removeLastSegmentFromPath(itemPath: string): string {
  * convertPathToMacStyle('smb:/path/to/folder'); // Returns 'smb://path/to/folder'
  */
 function convertPathToMacStyle(pathString: string): string {
-  if (pathString.startsWith('smb:/')) {
+  if (pathString.startsWith('smb://')) {
+    return pathString; // Already in Mac style
+  } else if (pathString.startsWith('smb:/')) {
     return pathString.replace('smb:/', 'smb://');
   }
   return pathString;
--- a/src/components/Help.tsx
+++ b/src/components/Help.tsx
@@ -1,9 +1,3 @@
-import React from 'react';
-
 export default function Help() {
-  return (
-    <div className="p-4">
-      <h2 className="text-foreground text-lg">Help Page</h2>
-    </div>
-  );
+  return <h2 className="text-foreground text-lg">Help Page</h2>;
 }
--- a/src/components/Jobs.tsx
+++ b/src/components/Jobs.tsx
@@ -7,38 +7,28 @@ export default function Jobs() {
   const { allTickets } = useTicketContext();
 
   return (
-    <div className="w-full flex justify-center">
-      <div className="w-full max-w-6xl p-6">
-        <Typography variant="h5" className="mb-6 text-foreground font-bold">
-          Jobs
-        </Typography>
-        <div className="rounded-lg shadow bg-background">
-          <div className="grid grid-cols-[2fr_3fr_1fr_1fr] gap-4 px-4 py-2 border-b border-surface">
-            <Typography variant="small" className="font-bold">
-              Target File
-            </Typography>
-            <Typography variant="small" className="font-bold">
-              Description
-            </Typography>
-            <Typography variant="small" className="font-bold">
-              Status
-            </Typography>
-            <Typography variant="small" className="font-bold">
-              Last Updated
-            </Typography>
-          </div>
-
-          {allTickets?.map(ticket => (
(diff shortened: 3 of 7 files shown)


src/utils/pathHandling.ts
```diff
@@ -141,7 +141,9 @@ function removeLastSegmentFromPath(itemPath: string): string {
  * convertPathToMacStyle('smb:/path/to/folder'); // Returns 'smb://path/to/folder'
  */
 function convertPathToMacStyle(pathString: string): string {
```
neomorphic, line 143: If this function always recieves a string that starts with smb, then you could simplify this to:
```typescript
function convertPathToMacStyle(pathString: string): string {
  return pathString.replace(/^smb:\/(?!\/)/, 'smb://');
}
```

allison-truhlar: The the POSIX join paths function normalizes all paths to use only single slashes, which is how the paths were ending up starting with `smb:/`. I just pushed a new commit that fixes this issue in a better way.

krokicki: Ah, I see. The smb mounts are not paths but URIs. We shouldn't use POSIX functions on them. Same goes for the windows "paths" which are actually Windows UNC path format. 

One way to deal with this consistently and safely is to always do operations (like concatenation, etc) on Linux-style paths, and convert to Windows/Mac only in the final display step.

allison-truhlar: The mount paths all start differently though, depending on whether they are Mac, Linux, or Windows and what storage type they're using. E.g.,

```
group: "scicompsoft"
linux_path: "/groups/scicompsoft/home"
mac_path: "smb://prfs.hhmi.org/scicompsoft$"
mount_path: "/groups/scicompsoft/home"
name: "groups_scicompsoft_home"
storage: "home"
windows_path: "\\\\prfs.hhmi.org\\scicompsoft$"
zone: "Scientific Computing Software"
```

What do you think of my approach in the [new commit](https://github.com/JaneliaSciComp/fileglancer/pull/114/commits/4116777d4c5641e5bca247e4235a008782fc2772#diff-32c0f22147c3c140629488a2984bf93282796d7b3b548ac56e17184af29daf1d)? I also added a [test](https://github.com/JaneliaSciComp/fileglancer/pull/114/commits/4116777d4c5641e5bca247e4235a008782fc2772#diff-905f3f8ef209996d234fd63d28fd14162306f4bf380e8d315b98c8e54f23222a) to confirm the specific case of smb:// being converted correctly.

krokicki: I think the issue here is that we're applying POSIX path operations to paths that are not in POSIX format. This breaks the assumptions those functions rely on, so the behavior becomes undefined. We're now dealing with some of the side effects, but it's both harder to fix retroactively and unclear whether we've caught every edge case.

The mount paths should always have the same format with no trailing slashes. We should normalize this during data loading in case any trailing slashes ever make their way into the wiki. Since we control the rest of the paths, they should be normalized to POSIX format upstream of the display function. Then the operation of forming the final display path can be very simple and predictable:
```
linux = mount_path + "/" + convert_to_linux(posix_subpath)
mac = mount_path  + "/" + convert_to_mac(posix_subpath)
windows = mount_path + "\\" + convert_to_windows(posix_subpath)
```

allison-truhlar: I just added a commit that I think follows the method you outlined, @krokicki. I ensure all the FSP mount paths (mount_path, linux_path, mac_path, and windows_path) don't have a trailing slash when they're loaded in the ZonesAndFspMap context. Then, I make sure the currentFileOrFolder `path` is normalized to have no leading slash and to use POSIX format (assuming it's already coming in in POSIX format - so this mostly just checks for no double slashes) inside the FileBrowserContext. Then display path works as follows:
```
linux = fsp[linux_path] + "/" + posix_subpath
mac = fsp[mac_path]  + "/" + posix_subpath
windows = fsp[windows_path] + "\\" + convert_to_windows(posix_subpath)
```

PR #115, opened 2025-07-21
Pre-push hook

allison-truhlar:
ClickUp id: 86aad2ztb
This PR is intended to help avoid some of the additional commits to update formatting using the Prettier rules, and some of the warning/error outputs in the GitHub build action. I used [Lefthook](https://github.com/evilmartians/lefthook)to set up the pre-push hook - I saw this recommended as a lightweight alternative to husky and I found it easy to set up and use. I chose to do a pre-push rather than pre-commit hook so local development and commits are slowed down.


krokicki (approved):
Very cool!

PR #117, opened 2025-07-23
Correct properties target state management

allison-truhlar:
ClickUp id: 86aaaybtb

This PR adds the propertiesTarget to the unified fileBrowserState in FileBrowserContext.tsx. Thiis fixes the problem of the properties target not being cleared if you navigate from a URL with FSP and/or file path params (e.g., `/browse/fsp_name/file_path`) to one without, (e.g. `/browse`). 

@krokicki I think this works - let me know if I should merge it or wait since we didn't plan to include it in the upcoming release.
@neomorphic


krokicki:
Works great, I'm going to merge it in for the release.

PR #118, opened 2025-07-24
Refactor and unify error handling

allison-truhlar:
@krokicki @neomorphic @cgoina 
This PR updates the error handling in the app to occur at the logic/UI boundary (i.e., custom hooks and functions returned by the context providers and called directly from the UI) and return a Result type. This result is then checked i the UI code (i.e., is result.success true or false), and this is used to determine what to display in the UI. For now, mostly toast alerts are displayed.

I also started adding some tests for components using React Testing Library and Mock Service Worker, which test that either toast.success or toast.error are called if the user completes an interaction that calls one of the functions that return a Result.

### DIFF
--- a/src/hooks/usePermissionsDialog.ts
+++ b/src/hooks/usePermissionsDialog.ts
@@ -85,7 +85,7 @@ export default function usePermissionsDialog() {
       if (response.ok) {
         return await refreshFiles();
       } else {
-        throw toHttpError(response);
+        throw await toHttpError(response);
       }
     } catch (error) {
       return handleError(error);
--- a/src/contexts/FileBrowserContext.tsx
+++ b/src/contexts/FileBrowserContext.tsx
@@ -172,7 +172,7 @@ export const FileBrowserContextProvider = ({
         } else if (response.status === 404) {
           throw new Error('Folder not found');
         } else {
-          throw toHttpError(response);
+          throw await toHttpError(response);
         }
       }
 
--- a/src/hooks/useDeleteDialog.ts
+++ b/src/hooks/useDeleteDialog.ts
@@ -27,7 +27,7 @@ export default function useDeleteDialog() {
         cookies['_xsrf']
       );
       if (!response.ok) {
-        throw toHttpError(response);
+        throw await toHttpError(response);
       } else {
         await refreshFiles();
         return createSuccess(undefined);
--- a/src/hooks/useRenameDialog.ts
+++ b/src/hooks/useRenameDialog.ts
@@ -37,7 +37,7 @@ export default function useRenameDialog() {
       if (response.ok) {
         return await refreshFiles();
       } else {
-        throw toHttpError(response);
+        throw await toHttpError(response);
       }
     } catch (error) {
       return handleError(error);


src/utils/errorHandling.ts
```diff
@@ -0,0 +1,32 @@
+import type { Success, Failure, Result } from '@/shared.types';
+
+function createSuccessResult<T>(data?: T): Success<T> {
```
krokicki, line 3: Since these functions will be used frequently, I would recommend choosing less verbose names to improve readability, such as `success()` and `failure()`.

src/utils/errorHandling.ts
```diff
@@ -0,0 +1,32 @@
...
+  return { success: true, data };
+}
+
+function createFailureResult(error: string): Failure {
+  return { success: false, error };
+}
+
+function parseError(error: unknown): string {
```
krokicki, line 11: This function doesn't do any parsing, so perhaps a better name for it is something like `getErrorMessage`.

src/hooks/useContextMenu.ts
```diff
@@ -73,18 +79,24 @@ export default function useContextMenu() {
...
+            folderPath: selectedFiles[0].path,
+            fsp: currentFileSharePath
+          },
+          'folder'
+        );
+        setShowContextMenu(false);
+      }
+    } catch (error) {
+      logger.error(error);
+      return createFailureResult(parseError(error));
```
krokicki, line 97: If the error handling pattern will always compose these functions, then should we have a function for it? Something like `return handleError(error)` and then that function would log, get the message, and return the failure object.

src/utils/errorHandling.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  if (error instanceof Error) {
+    return error.message;
+  }
+
+  return 'An error occurred';
+}
+
+function checkIfSuccessful<T>(result: Result<T>) {
```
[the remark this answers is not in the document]
krokicki: I think Jody's right. I was thinking of a method like you would have in Java, but on second thought it's not necessary here.

src/components/ui/Menus/ContextMenu.tsx
```diff
@@ -36,14 +42,26 @@ export default function ContextMenu({
...
+  const { folderPreferenceMap } = usePreferencesContext();
+
+  const isFavorite: boolean = folderPreferenceMap[
+    makeMapKey(
+      'folder',
+      `${fileBrowserState.currentFileSharePath?.name}_${selectedFiles[0].path}`
+    )
+  ]
+    ? true
+    : false;
```
krokicki, line 63: A more idiomatic way of doing this conversion would be to either use the `!!` (double bang) operator, or the `Boolean()` constructor.

src/hooks/usePermissionsDialog.ts
```diff
@@ -1,43 +1,86 @@
...
+        return prev; // Ensure the type remains consistent
+      }
+      const splitPermissions = prev.split('');
+      if (checked) {
+        splitPermissions.splice(parseInt(position), 1, value);
+      } else {
+        splitPermissions.splice(parseInt(position), 1, '-');
+      }
+      const newPermissions = splitPermissions.join('');
+      return newPermissions;
```
krokicki, line 40: This code would benefit from additional comments and unit tests.

src/shared.types.ts
```diff
@@ -27,15 +27,30 @@ type ZonesAndFileSharePathsMap = Record<string, FileSharePath | Zone>;
...
+interface Failure {
+  success: false;
+  error: string;
+}
+
+interface ApiFailure extends Failure {
+  code: number;
+}
+
+type Result<T> = Success<T> | Failure | ApiFailure;
```
krokicki, line 44: When using discriminated unions, it's best to use types which do not overlap so that the type system can discriminate cleanly. In this case, `ApiFailure` is also a `Failure`, so they are overlapping and will never be discriminated. I would separate these into separate types and add a discriminator string to tell them apart. This is assuming we need `ApiFailure`, which I'm not really sure about (see other comment).

src/utils/errorHandling.ts
```diff
@@ -0,0 +1,49 @@
...
+async function getResponseError(response: Response): Promise<string> {
+  const body = await response.json();
+  return body.error ? body.error : 'Unknown error';
+}
+
+// Adding the option to return a ApiFailure Result type rather than
+// throwing an error because different functions might want to
+// handle response status codes differently. E.g., 404 might be an
+// error sometimes, but for preferences, it just means that preference
+// key hasn't been set yet.
```
krokicki, line 17: Could this be handled with exceptions in the business logic? I'm not sure this kind of decision needs to be bubbled up to the UI. Is there an example?

src/shared.types.ts
```diff
@@ -27,15 +27,30 @@ type ZonesAndFileSharePathsMap = Record<string, FileSharePath | Zone>;
 
 type Cookies = { [key: string]: string };
 
-type Result<T, E extends Error> =
-  | { ok: true; value: T }
-  | { ok: false; error: E };
+type Success<T> = {
+  success: true;
+  data?: T;
```
krokicki, line 32: Since the data is optional here, the type system can't help us with validating when the data can be missing. I suggest using a `Result<void>` type when the function doesn't need to return data, and returning `data: undefined` when there is a success.

src/utils/copyText.ts
```diff
@@ -23,19 +25,21 @@ function fallbackCopyTextToClipboard(text: string) {
   }
 }
 
-const copyToClipboard = async (text: string | null) => {
+const copyToClipboard = async (text: string | null): Promise<Result<null>> => {
```
krokicki, line 28: I think this (and other similar functions) should be `Result<void>`. This would indicate that there is no data ever produced by these functions. `Result<null>` is similar but it implies that there happens to be no data for a result (but there could be). This is useful for when you are getting a value but there may not be one e.g. `Result<string | null>`.

src/shared.types.ts
allison-truhlar: An example would be in TicketsContext, where there is a function that is called in a useEffect that fires when the app first mounts. For example, [fetchTicket](https://github.com/JaneliaSciComp/fileglancer/blob/47db99fbccc65aed2d653bc98e0004c9a7dc79fb/src/contexts/TicketsContext.tsx#L75) in TicketsContext is called in [this useEffect](https://github.com/JaneliaSciComp/fileglancer/blob/47db99fbccc65aed2d653bc98e0004c9a7dc79fb/src/contexts/TicketsContext.tsx#L149) which then results in a 404 error being logged to the console. If the status code was preserved, as in the ApiFailure type, then it could be checked in the useEffect and used to print the error using log.debug instead, or another less intrusive way of logging the result. And if fetchTicket is called from logic related to the UI, then the ApiFailure result can be used in the same way as other results to get an error string to display to the user.

allison-truhlar: Removed ApiFailure type, commit https://github.com/JaneliaSciComp/fileglancer/pull/118/commits/873113cabad2204f54640a37a262cc63eb489efd

src/shared.types.ts
allison-truhlar: Require data in Success type and pass `data: undefined` when there is a success, commit https://github.com/JaneliaSciComp/fileglancer/pull/118/commits/146866ea03aaa809bc98f2b0f2cf83e3199a1d95

src/utils/errorHandling.ts
krokicki: I think the underlying issue is that `fetchTicket` is missing a return type. It returns a Result when there is an error, but a bare object when there is success. If we set the return type to `Promise<Result<Ticket | null>>`, I think the type system would force solutions to a lot of problems, which is our goal with this refactoring. For example, it would be clear that this function must be responsible for parsing the HTTP return code and deciding what/how to log and what to return to the user. Also, the `useEffect` (hopefully) won't need to try/catch.

src/contexts/TicketsContext.tsx
```diff
@@ -209,7 +173,11 @@ export const TicketProvider = ({ children }: { children: React.ReactNode }) => {
         log.error('Error in useEffect:', error);
       }
     })();
-  }, [fetchTicket, propertiesTarget]);
+  }, [
+    fetchTicket,
+    fileBrowserState.propertiesTarget,
+    fileBrowserState.currentFileSharePath
```
krokicki, line 179: `fetchTicket` already has these 2 dependencies, so they are redundant here and may cause this to run more often than necessary.

allison-truhlar: Fixed in this commit: https://github.com/JaneliaSciComp/fileglancer/pull/118/commits/aebba7c255fc426d4210c5120fbf58d3d748de94

src/utils/errorHandling.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (error instanceof Error) {
+    return error.message;
+  }
+  if (error instanceof Response) {
+    return await getResponseError(error);
+  }
+  return 'An unknown error occurred';
+}
+
+async function handleError(error: unknown): Promise<Failure> {
```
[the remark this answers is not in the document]
allison-truhlar: This makes sense - I did think it was probably a bad idea to add so many "awaits" to the code. I'll refactor following your suggestion.

krokicki (approved):
Looks good, thanks for all your work on this Allison!

PR #119, opened 2025-07-24
feat: add content to help page, including version number

allison-truhlar:
ClickUp id: 86aafd3hn

This PR logic to fetch the Fileglancer version number from the API and display it in the navbar.
It also creates a common <Tag> component for use by both the version number tag and "BETA" tag in the navbar.
@krokicki @neomorphic


allison-truhlar:
I made Konrad's requested changes - the help page is functional, but I welcome any feedback on the style/layout.

neomorphic (approved):
I think the overall style of the help page is fine. I would just change the way the links are displayed to more closely resemble the links on the other pages of the site. I might also display the version number more prominently, with a string like "Version X.X.X", so that people have an easier time locating it when they are frustrated and want to report a bug. I don't suppose it is possible to add the version number to the form that is submitted to click up?

krokicki:
@neomorphic ClickUp does support [prefilling values with URL parameters](https://forms.clickup.com/10502797/f/a0gmd-713/NBUCBCIN78SI2BE71G?Version=1.0.0), but I can't figure out how to get it to work. It's possible that it requires a higher subscription plan.

allison-truhlar:
@neomorphic I made the color change you suggested for the version number.

For the `FgCard` component - my intention with creating it using the `<Card`> component from material-tailwind and adding some extra styles is so that I could ensure these were consistent across all pages, and if I change the style for one, I change it for all. Looking at the [base styles](https://github.com/creativetimofficial/material-tailwind/blob/main/packages/material-tailwind-react/src/theme/components/card/index.ts) for the card component, I can whittle down these extra styles to just `border-surface dark:border-foreground`, to ensure the border matches the row borders used in the Jobs and Data Links cards.

PR #120, opened 2025-07-29
Dashboard UI

allison-truhlar:
ClickUp id: 86aad58pe

Adds a dashboard display for when /fg/browse is first loaded. Includes:
- Card with list of 10 most-recently viewed folders (data stored in a new `recentlyViewedFolders` preference key). This has been updated to ensure that, when navigating through the directory tree, only the last folder visited on a given path is saved to preferences.
- Card with recently created data links

Also adds a button to the /browse toolbar that takes the user to their home directory in the file browser. The data for the home directory file share path name and base name are obtained from new methods in the backend, in the `BaseHandler` and `ProfileHandler`.

Finally, adds a `<FgTooltip/>` component and refactors all tooltips in the app to use this code.

@krokicki @neomorphic


fileglancer/handlers.py
```diff
@@ -39,6 +39,29 @@ def get_current_user(self):
...
+    def get_home_file_share_path_name(self):
+        """
+        Get the file share path for the current user's home directory.
+        
+        Returns:
+            str: The file share path name for the user's home directory, formatted as file_share_path_name
+        """
+        home_file_share_path = os.path.split(self.get_home_directory_path())[0]
+        file_share_paths = get_fsp_manager(self.settings).get_file_share_paths()
+        for fsp in file_share_paths:
```
neomorphic, line 60: Was there a discussion of caching the result from this loop, so that it only had to be performed once? Is the caching done somewhere else in the code?

src/components/ui/BrowsePage/Dashboard/RecentDataLinksCard.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  // Get the 10 most recent data links
+  const recentDataLinks = allProxiedPaths?.slice(0, 10) || [];
+
+  return (
+    <DashboardCard title="Recently created data links">
+      {recentDataLinks.length === 0 ? (
+        <div className="text-center py-8">
+          <Typography variant="small" className="text-muted-foreground">
+            No data links created yet
```
neomorphic, line 19: This is going to be one of the first things that people see when they access the site. It would probably be good to add more information about how to create a data link or where they can go look to get more information on the topic.

src/layouts/MainLayout.tsx
```diff
@@ -23,8 +23,8 @@ export const MainLayout = () => {
   return (
     <CookiesProvider>
       <ZonesAndFspMapContextProvider>
-        <FileBrowserContextProvider fspName={fspName} filePath={filePath}>
-          <OpenFavoritesProvider>
+        <OpenFavoritesProvider>
```
neomorphic, line 26: Did AI swap the order of these providers, or was there some underlying reason why they needed to be nested in the reverse order?

src/components/ui/BrowsePage/DataToolLinks.tsx
```diff
@@ -0,0 +1,117 @@
...
+}): React.ReactNode {
+  const [showCopiedTooltip, setShowCopiedTooltip] = React.useState(false);
+
+  const handleCopyUrl = async () => {
+    if (urls?.copy) {
+      await copyToClipboard(urls.copy);
+      setShowCopiedTooltip(true);
+      setTimeout(() => {
+        setShowCopiedTooltip(false);
+      }, 1000);
```
neomorphic, line 28: Usually when I set these alerts to one second, Konrad asks me to extend the timeout. Maybe 2 - 3 seconds.

src/components/ui/widgets/FgTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+  label,
+  triggerClasses,
+  openCondition,
+  children
+}: FgTooltipProps) {
+  const Component = as || null;
+  const Icon = icon || null;
+
+  return (
+    <Tooltip placement="top" open={openCondition}>
```
neomorphic, line 32: I am assuming you haven't need to place a tooltip in any other placement up to this point. Since that is the case, setting top is fine, but the first time we need to place the tooltip in a different placement, we should make this an option with top as the default.

src/contexts/PreferencesContext.tsx
```diff
@@ -339,6 +345,52 @@ export const PreferencesProvider = ({
     ]
   );
 
+  const updateRecentlyViewedFolders = React.useCallback(
```
neomorphic, line 348: This function largely works as I would have expected, except for when I navigate from a groups home directory to a sub directory, eg: navigating from `/groups/cellmap/cellmap` to ` /groups/cellmap/cellmap/db_backup` results in two entires in the recently viewed table. I believe this is because the `/groups/cellmap/cellmap` has a folderPath value of '.', which does not match the start of the child paths folderPath value.

neomorphic (approved):
Looks good. I added a few comments.

src/components/ui/BrowsePage/Dashboard/RecentDataLinksCard.tsx
allison-truhlar: I slightly modified some text from the Data Links page and added it here when there are no data links created yet. https://github.com/JaneliaSciComp/fileglancer/pull/120/commits/08a7b936fc28e1fd565454b861bada4826f7c420#diff-4f11b05ca589d9640f1ee1362e25c937c73919203d7fc5a13fe286bf090fbe8fL15

src/contexts/PreferencesContext.tsx
allison-truhlar: Based on our conversation, I removed file share paths (ie. folderPath = '.') from recentlyViewedFolders (commit https://github.com/JaneliaSciComp/fileglancer/pull/120/commits/fffbf36648db5ec15b213f2413d14b981f4b00f2). The most motivating factor is that if the file share path is only saved in the case that you navigate to a file share path and then navigate away without visiting any descendant items, then this is likely saving a not very useful to you item. In addition, you can already click to access all the file share paths in the sidebar, which you can't do for folders.

PR #121, opened 2025-08-06
External data links

krokicki:
This uses the new API provided in https://github.com/JaneliaSciComp/fileglancer-central/pull/23 to check for external buckets in the current path. If found, it generates external links alongside the internal ones. The UI display still needs some work.

@StephanPreibisch @cgoina

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@ node_modules
 fileglancer
 .github
 /style/tailwind_converted.css
+.claude/settings.local.json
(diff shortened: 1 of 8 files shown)


allison-truhlar (approved):
Looks good! I added a few refactoring commits:

- I refactored ExternalBucketContext to use fileBrowserState instead of the individual currentFSP and currentFolder states. I also added some checks in updateExternalBucket for whether these states exist to avoid using the "optional" notation (e.g., fileBrowserState.currentFileSharePath?.name) in the function when these aren't optional.
- I refactored the two types of links out into a new reusable component, DataToolLinks.

PR #128, opened 2025-08-18
feat: save side panel state using the preference API

allison-truhlar:
@krokicki @neomorphic 
Clickup id: 86aad9vk0
This PR saves the side panel state in the database using the existing preference API. To do this, I added a custom storage API for the side panels made using `react-resizable-panels`. Per the [docs,](https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/README.md#panelgroup) the custom storage API exposes `getItem()` and `setItem()` synchronous methods. The `setItem()` method ultimately calls `handleUpdateLayout`, which makes the call to the preferences API. The `getItem()` method attempts to set the layout using the layout state, which comes from `PreferencesContext.tsx`. The layout is in JSON format, and includes a layout for both a three-panel set up (key = `main, properties, sidebar`) and a two-panel set up when the properties drawer is collapsed (key = `main, properties`). This is why the `showPropertiesDrawer` state is now contained within the `useLayoutPrefs.ts` hook - it toggles which key to use to lookup the layout to apply.


src/components/Browse.tsx
```diff
@@ -17,7 +17,7 @@ import RecentlyViewedCard from './ui/BrowsePage/Dashboard/RecentlyViewedCard';
 
 type OutletContextType = {
   setShowPermissionsDialog: React.Dispatch<React.SetStateAction<boolean>>;
-  setShowPropertiesDrawer: React.Dispatch<React.SetStateAction<boolean>>;
+  togglePropertiesDrawer: () => void;
```
neomorphic, line 20: This type could be defined in BrowseLayout.tsx and then imported so that the data structure doesn't have to be defined in two different places. eg:

  // In BrowseLayout.tsx
  export const outletContextValue = { /* ... */ };
  export type OutletContextType = typeof outletContextValue;

  // In Browse.tsx
  import type { OutletContextType } from '@/layouts/BrowseLayout';
  const context = useOutletContext<OutletContextType>();

src/hooks/useLayoutPrefs.ts
```diff
@@ -0,0 +1,156 @@
+import React from 'react';
+
+import { usePreferencesContext } from '@/contexts/PreferencesContext';
+import logger from '@/logger';
+
+/**
+ * Custom hook that provides storage interface for react-resizable-panels
```
neomorphic, line 7: When the show properties drawer is closed and the left panel is resized, the properties drawer springs open again.

neomorphic: Closing the properties drawer and refreshing the page causes the drawer to open again

src/hooks/useLayoutPrefs.ts
```diff
@@ -0,0 +1,156 @@
...
+    },
+    [handleUpdateLayout]
+  );
+
+  const togglePropertiesDrawer = () => {
+    setShowPropertiesDrawer(prev => !prev);
+  };
+
+  // Initialize layouts from saved preferences
+  React.useEffect(() => {
```
neomorphic, line 47: Without looking too deeply, my guess is a useEffect is not triggering when it should, causing the properties drawer state to be set incorrectly.

src/components/Browse.tsx
allison-truhlar: I couldn't do quite this, because TypeScript wasn't letting me put an export inside the BrowseLayout function, but I did move the definition for the OutletContextType to BrowseLayout.tsx, to keep it next to where the value is actually used. I also exported it for use in Browse.tsx, as suggested. 
https://github.com/JaneliaSciComp/fileglancer/blob/d9c520f546f53b1b464b6d504bd1cd36ee46ace6/src/layouts/BrowseLayout.tsx#L11
See full changes in this commit: https://github.com/JaneliaSciComp/fileglancer/pull/128/commits/d9c520f546f53b1b464b6d504bd1cd36ee46ace6

src/hooks/useLayoutPrefs.ts
allison-truhlar: @neomorphic Both of these bugs should be fixed now - can you retest? I also realized I hadn't accounted for the case where the sidebar is closed. This is also incorporated now.

neomorphic: @allison-truhlar I have checked out the latest changes and that seems to have fixed the issues

neomorphic (approved):

PR #130, opened 2025-08-25
feat: add sorting and resizable cols to file browser

allison-truhlar:
Clickup id: 86a83e1bg

This PR uses Tanstack Table (formerly React Table) to create a file table with sorting and resizable columns. 

As part of this process, I refactored selectedFiles state from a hook that was only storing this state into the unified fileBrowserState stored in FileBrowserContext. I also moved the logic related to updating the selectedFiles when you left or right click into FileBrowserContext, since this is where selectedFiles and propertiesTarget already are. These actions allowed me to get rid of the setPropertiesTarget state setter that was being exposed from FileBrowserContext. In addition, I moved handleContextMenuFavorite out of the useContextMenu hook into the PreferencesContext, since this is where the FSP and folder favorites are stored. Now the useContextMenu hook just deals with positioning the context menu.

### DIFF
--- a/src/hooks/useZarrMetadata.ts
+++ b/src/hooks/useZarrMetadata.ts
@@ -123,7 +123,13 @@ export default function useZarrMetadata() {
         }
       }
     },
-    [areFileDataLoading, fileBrowserState, cookies]
+    [
+      areFileDataLoading,
+      fileBrowserState.currentFileSharePath,
+      fileBrowserState.currentFolder,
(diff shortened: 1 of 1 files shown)


krokicki:
This works nicely!

I would merge this, but afterwards it might be a good time to go back and refactor the `updateAllStates` method into the set of state transitions that @neomorphic proposed. We should create a state diagram first to organize things. Happy to discuss more.

PR #131, opened 2025-08-26
feat: add hide dot files to user preferences

allison-truhlar:
Clickup id: 86aa8dhm4

This PR moves the hideDotFiles state from the useHideDotFiles hook to PreferencesContext, and adds logic in PreferencesContext to save the hideDotFiles state to the backend. Now the useHideDotFiles hook is only used to calculate the files that should be displayed in the UI, based on the current files in fileBrowserState and whether hideDotFiles is true/false.

For the UI, a section is added to the Preferences page where the user can click a checkbox to toggle whether dot files are hidden or not. I also updated the button in the toolbar that toggles hide dot files to show toast success/error messages, like the Preferences page.

### DIFF
--- a/src/components/Preferences.tsx
+++ b/src/components/Preferences.tsx
@@ -18,96 +18,100 @@ export default function Preferences() {
         Preferences
       </Typography>
 
-        <Card>
-          <Card.Header>
-            <Typography className="font-semibold">
-              Format to use for file paths:
+      <Card>
+        <Card.Header>
+          <Typography className="font-semibold">
+            Format to use for file paths:
+          </Typography>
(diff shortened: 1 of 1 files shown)


krokicki (changes requested):
This works, but the UX is now inconsistent. The path format preferences requires the user to press Submit, but the checkbox saves automatically. We should pick one way and go with it for everything. Either one save button at the bottom that saves all preference changes which were made, or every preference saves itself automatically when changed.

krokicki:
I went ahead and removed the Submit button so that we can merge this, because I want to work on adding more more preferences.

PR #137, opened 2025-08-28
Consistent height for navbar

krokicki:
This fix (by Claude Code) makes the top navbar's height consistent when clicking across the various sections (Browse, Data Links, etc.) I also tested that it works as expected at the narrower "mobile" widths. 

@StephanPreibisch @neomorphic

### DIFF
--- a/src/components/ui/Navbar/Navbar.tsx
+++ b/src/components/ui/Navbar/Navbar.tsx
@@ -87,7 +87,7 @@ export default function FileglancerNavbar() {
   }, [setIsLightTheme]);
 
   return (
-    <Navbar className="mx-auto w-full rounded-none bg-background p-4 short:py-1 short:px-1 dark:shadow-surface min-h-[45px]">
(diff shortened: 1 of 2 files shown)


allison-truhlar (approved):
@krokicki This mostly worked for me - I added several commits for these issues I was experiencing:

- Fixed the main layout so that the content on Data Links, Tasks, and Help was centered again. The added divs caused the content to be left-aligned.

On cell phone size screens:
- The text in the logo wrapped and the bottom of "Fileglancer" got cutoff. I reduced the padding so that the default is now p-2, and only once reaching the "sm" cutoff (min width 640px), the padding increases to p-4.
- The height of the Navbar changed when toggling the menu open and closed, since it was set to a min-h-[61px] when the menu was open (and could therefore get larger), and to a set height of 61px when the menu was not open. I changed this so that it is now always a height of 61px. 
- Ensure the content of the Navbar is always horizontally centered by adding 'flex items-center`.
- Due to adding the flex class on Navbar, I moved the `<Collapse>` component for the menu outside of the `<Navbar>` component. I then had to add a background color that matched the navbar.

PR #139, opened 2025-08-31
Metadata table improvements

krokicki:
Clickup id: 86ab01anm

Refactored axes metadata into secondary table based on feedback from @mkitti .

The metadata for an OME-Zarr now looks like this:
Screenshot 2025-08-31 at 12 48 22 PM

While a bare Zarr array looks like this:
Screenshot 2025-08-31 at 12 48 28 PM

This PR also fixes a bug that was preventing the "open with" links from showing up.

@StephanPreibisch @allison-truhlar @neomorphic


mkitti:
The order of the axes does appear strange. I guess it is using the C / Python order, but this seems backwards to a human.

PR #143, opened 2025-09-03
Simple file viewer

krokicki:
Clickup id: 86aa5bzbu

This PR adds a simple file viewer which shows common file types with syntax highlighting. This is mainly intended for viewing JSON files associated with OME-Zarr, but can also be used to view other text files on the file system. 

Screenshot 2025-09-02 at 8 23 27 PM

As part of this implementation, Claude Code decided to refactor some of the deprecated file browser states, so I went with it and tried to fix it everywhere. As part of *that*, I ended up updating Typescript and adding aliases for type checking, because it needed to see the same errors I was seeing in VS Code on the command line. There are now some additional commands for this:

```
pixi run node-check # runs type checking
pixi run node-lint # runs linting
pixi run node-lint-write # fix linting
```

But I've realized that this needs a bigger overhaul to clean up and normalize commands among `pyproject.yaml`, `package.json`, GitHub Actions, and lefthook, but we can do that in a separate PR.

@StephanPreibisch


neomorphic (approved):
Looks good to me

PR #148, opened 2025-09-05
dev: fix pre-push hook

allison-truhlar:
Clickup id: 86abdq8h6

Previously, when you pushed and the Prettier pre-check made formatting changes, it left those as uncommitted in the local repo and pushed as-is. Now, it rejects the push. The pre-check now uses a shell script that also calls `prettier --write`, for convenience, if the Prettier check fails. The user then has to commit these changes and try the push again.

In the course of making these edits, I started consolidating more commands into pixi. I moved all the linting commands into the pixi command list, only, as opposed to having them replicated in both the pixi command list, the package.json command list, and the GH actions `build.yml` file.

I added some comments in the file to explain why other scripts, namely `clean` and `install:extension`, were removed from the package.json file.

@krokicki @neomorphic


package.json
```diff
@@ -29,19 +29,13 @@
         "build:prod": "tsc -p tsconfig.extension.json && vite build --mode production && npm run build:labextension",
         "build:labextension": "jupyter labextension build .",
         "build:labextension:dev": "jupyter labextension build --development True .",
-        "clean": "npm run clean:lib",
```
allison-truhlar, line 32: I removed this script because it was not called anywhere.

package.json
```diff
@@ -29,19 +29,13 @@
...
         "clean:labextension": "rimraf fileglancer/labextension fileglancer/_version.py",
         "clean:all": "npm run clean:lib && npm run clean:labextension && npm run clean:lintcache",
         "watch": "run-p watch:src watch:labextension",
         "watch:src": "vite build --watch",
         "watch:labextension": "jupyter labextension watch .",
-        "lint": "eslint .",
-        "eslint:check": "eslint . --cache --ext .ts,.tsx",
-        "prettier:check": "npx prettier \"**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}\" --check",
-        "lint:check": "npm run prettier:check && npm run eslint:check",
-        "install:extension": "npm run build",
```
allison-truhlar, line 44: I removed the `install:extension` script because it just calls the `build` script. I then replaced the one usage of `install:extension`, found in line 91 of pyproject.toml, with `build`.

pyproject.toml
```diff
@@ -88,7 +88,7 @@ skip-if-exists = ["fileglancer/labextension/static/style.js"]
 build_cmd = "build:prod"
 
 [tool.hatch.build.hooks.jupyter-builder.editable-build-kwargs]
-build_cmd = "install:extension"
```
allison-truhlar, line 91: I removed the `install:extension` script because it just calls the build script. I then replaced this one usage of `install:extension` with `build`.

PR #151, opened 2025-09-08
feat: data link reverse mapping

allison-truhlar:
Clickup id: 86abc1r9t

@krokicki @neomorphic 
This PR follows the data link reverse mapping method suggested by Jody - adding the sharing key as a column in the data links table. In this process, I decided to convert the `<TableCard>` component, which wraps the tables on both the `/links` and `/jobs` pages, to use Tanstack Table. Now, you can sort both tables by certain columns and I also added a search input to the sharing key column on the `/links` page. 

https://github.com/user-attachments/assets/bc5ca19f-ae3e-4ecc-ad47-015aae40587a

Another motivation to switch to Tanstack Table is that the [docs](https://tanstack.com/table/latest/docs/guide/pagination#should-you-use-client-side-pagination) say it is able to handle up to 10s of 1000s of entries for client-side pagination. I will implement this with a separate PR/Clickup task.


src/components/ui/Table/linksColumns.tsx
```diff
@@ -146,6 +136,87 @@ export default function ProxiedPathRow({ item }: ProxiedPathRowProps) {
...
+          >
+            <Typography className="text-foreground truncate">
+              {item.sharing_name}
+            </Typography>
+          </FgTooltip>
+        </div>
+      );
+    },
+    enableSorting: true,
+    enableColumnFilter: false
```
neomorphic, line 165: I noticed that you didn't have to set the `enableColumnFilter: false` on the jobsColumns object. Does the false have to be set here, because you set one of the columns to true?

src/components/Links.tsx
```diff
@@ -1,8 +1,8 @@
 import { Typography } from '@material-tailwind/react';
 
 import { useProxiedPathContext } from '@/contexts/ProxiedPathContext';
-import ProxiedPathRow from './ui/LinksPage/ProxiedPathRow';
-import { TableCard } from './ui/widgets/TableCard';
+import { TableCard } from './ui/Table/TableCard';
+import { linksColumns } from './ui/Table/linksColumns';
```
neomorphic, line 5: Should these be using the `@/components/ui/Table/linksColumns` imports?

allison-truhlar:
Thanks, @krokicki! I added a filter function so now it works on the formatted date string (which is what is displayed), as opposed to the raw date string.

PR #152, opened 2025-09-09
Neuroglancer state generation improvements

krokicki:
Clickup ids: 86abk3vnm and 86abn3x6p

When Neuroglancer state generation is disabled, this now generates the same state that you would get if you pasted the URL directly into Neuroglancer.

In addition, we now implement a sampling step to classify the image type automatically. It randomly samples a subset of voxels from the volume and compute the ratio of unique intensity values to the total number of samples. If this ratio exceeds a predefined threshold (currently 0.1%), infer that the dataset represents a segmentation volume rather than a natural image. In such cases, we automatically configure Neuroglancer with the "segmentation" layer setting. 

@StephanPreibisch @vaxenburg @allison-truhlar @neomorphic

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Test the extension frontend
       run: |
         set -eux
-        pixi run --environment default test-frontend
+        pixi run test-frontend
 
     - name: Test the extension backend
       run: |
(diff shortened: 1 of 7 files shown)


vaxenburg:
I'm curious how the layer type inference works with the 0.1% threshold and sampling from the entire chunk. I thought it was important to count unique values in close proximity (for example all unique values in a bunch of small 10x10x10 cubes) as indication of how fast the values change locally. But I guess it works fine when you sample from the entire chunk too?

Also, I think the 0.1% threshold means there are less than 10,000 * 0.001 == 10 unique labels in the chunk. Isn't it a bit too small a threshold? But yeah, you probably tested all that..

krokicki:
@vaxenburg I tested it on some natural images and a few of your segmentations. You make some good points. I agree that counting adjacent values should be a better discriminator, so I made that change. The threshold is something that we will need to tune as we see more data, I only have a few segmentations to test with. I also added a preference to disable this heuristical behavior in case it's an issue.

vaxenburg:
Nice, thanks. If it helps, I can try testing it on a larger amount of data maybe..

Optional disabling of the heuristic is a good idea.

PR #156, opened 2025-09-10
feat: add preference to enable automatic data link creation

allison-truhlar:
Clickup id: 86aata5u9

This PR adds a user preference to enable automatic data link creation. This preference can be changed on the preference page, or from the data link dialog. The data link toggle switch is removed from Zarr directories. Instead, the data viewer icons are shown immediately. The first time a user clicks on an icon to make a data link, a modal appears offering options to "create link" or "cancel" and a checkbox to enable automatic data links going forward (this is the same checkbox as on the preference page). Once automatic data links are enabled, the user does not see this modal, and when they navigate to Zarr directory, the data link is automatically created, so when they click on a data viewer icon, they are immediately taken to the data viewer.

https://github.com/user-attachments/assets/4dde0df2-deb3-4db0-a97d-d60eba6dd601

I also refactored some styles, specifically, the custom colors. The "error" colors were not providing sufficient contrast for the "delete" and "create data link" buttons when the user hovered on them.

@krokicki @neomorphic

### DIFF
--- a/src/components/ui/Dialogs/DataLink.tsx
+++ b/src/components/ui/Dialogs/DataLink.tsx
@@ -24,6 +24,153 @@ type DataLinkDialogProps = {
   setPendingNavigationUrl?: React.Dispatch<React.SetStateAction<string | null>>;
 };
 
+function CreateLinkBtn({
+  displayPath,
+  pendingNavigationUrl,
+  setPendingNavigationUrl,
+  setShowDataLinkDialog
+}: {
+  displayPath: string;
+  pendingNavigationUrl?: string | null;
+  setPendingNavigationUrl?: React.Dispatch<React.SetStateAction<string | null>>;
+  setShowDataLinkDialog: React.Dispatch<React.SetStateAction<boolean>>;
+}): JSX.Element {
+  const { createProxiedPath, fetchProxiedPath, refreshProxiedPaths } =
+    useProxiedPathContext();
+
+  async function refreshLinks() {
+    const refreshResult = await refreshProxiedPaths();
+    if (!refreshResult.success) {
+      toast.error(`Error refreshing proxied paths: ${refreshResult.error}`);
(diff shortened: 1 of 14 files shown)


neomorphic (approved):
LGTM

vite.config.ts
```diff
@@ -41,6 +41,7 @@ export default defineConfig({
         '**/dist/**',
         '**/ui-tests/**'
       ]
-    }
+    },
+    silent: 'passed-only'
```
allison-truhlar, line 45: Adding this parameter address Clickup id 86abn3rw3. Now, only the logs for failing tests are printed, to help with debugging.

src/components/ui/Dialogs/DataLink.tsx
```diff
@@ -2,39 +2,155 @@ import React from 'react';
...
+  pendingToolUrl?: PendingToolUrl;
+  setPendingToolUrl?: React.Dispatch<React.SetStateAction<PendingToolUrl>>;
+  handleCopyUrl?: (url: string) => Promise<void>;
+  handleCreateDataLink?: (
+    pendingToolUrl: PendingToolUrl,
+    urls: OpenWithToolUrls | null,
+    handleCopyUrl: ((url: string) => Promise<void>) | undefined,
+    setPendingToolUrl:
+      | React.Dispatch<React.SetStateAction<PendingToolUrl>>
+      | undefined
```
krokicki, line 32: When can `setPendingToolUrl` be undefined?

krokicki (changes requested):
I found one issue: OME-Zarrs are not being correctly opened in Neuroglancer because the state is being generated using only part of the logic in useZarrMetadata. This only affects the first time Neuroglancer is opened after automatic link generation. 

Besides that, there's a few comments and edits I'm making for clarity.

PR #157, opened 2025-09-10
Support Zarr v3 and  OME-Zarr 0.5

krokicki:
Clickup id: 86abnxvv0

This PR adds support for Zarr v3 and OME-Zarr 0.5. A couple major backend changes were necessary in order to read from sharded chunks:
* Implemented HTTP Range reads in [this x2s3 PR](https://github.com/JaneliaSciComp/x2s3/pull/11)
* Implemented HTTP Range reads in the Fileglancer user server (currently just used for the thumbnail generation)

The frontend was refactored to check for `.zarray` first, then `.zattrs`, and if neither exists it checks for a `zarr.json` and looks at the `node_type` to see if it's an array or group. 

@mkitti @neomorphic @allison-truhlar @dchen116  @StephanPreibisch

### DIFF
--- a/src/hooks/useZarrMetadata.ts
+++ b/src/hooks/useZarrMetadata.ts
@@ -120,9 +120,12 @@ export default function useZarrMetadata() {
     }
   };
 
-  const getFile = async (fileName: string) => {
-    return fileBrowserState.files.find(file => file.name === fileName);
-  };
+  const getFile = React.useCallback(
+    async (fileName: string) => {
+      return fileBrowserState.files.find(file => file.name === fileName);
+    },
+    [fileBrowserState.files]
+  );
 
   const checkZarrMetadata = React.useCallback(
     async (cancelRef: { cancel: boolean }) => {
@@ -188,7 +191,7 @@ export default function useZarrMetadata() {
       areFileDataLoading,
       fileBrowserState.currentFileSharePath,
       fileBrowserState.currentFileOrFolder,
-      fileBrowserState.files,
+      getFile,
       cookies
     ]
   );


src/hooks/useZarrMetadata.ts
```diff
@@ -77,70 +145,43 @@ export default function useZarrMetadata() {
           fileBrowserState.currentFileSharePath.name,
           fileBrowserState.currentFileOrFolder.path
         );
-        const zarrayFile = fileBrowserState.files.find(
-          file => file.name === '.zarray'
-        );
+
+        const zarrayFile = await getFile('.zarray');
         if (zarrayFile) {
```
mkitti, line 150: It looks like we iterate through `fileBrowserState.files` a few times to look for three files. Perhaps it would make sense to accumulate these into a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) once. We can then check the `Set` for the files via `has`.

The exact implementation is not specified other than the time complexity being better than `O(n)`, which is where we currently are.

src/omezarr-helper.ts
```diff
@@ -432,9 +432,11 @@ function generateNeuroglancerStateForOmeZarr(
...
 
-async function getZarrArray(dataUrl: string): Promise<zarr.Array<any>> {
+async function getZarrArray(
+  dataUrl: string,
+  zarrVersion: 2 | 3
+): Promise<zarr.Array<any>> {
   log.debug('Getting Zarr array for', dataUrl);
-  const zarrVersion = 2;
   const store = new zarr.FetchStore(dataUrl);
   return await omezarr.getArray(store, '/', zarrVersion);
```
mkitti, line 441: This pre-exists this pull request, but a dimension separator of `.` is technically the default Zarr v2 separator. We have most often seen this from Zarr arrays sent to us from outside contractors such as Ariadne. I wonder if too many low-level Zarr details are passing through the abstraction barrier here.

Perhaps we should just detect the presence of a `.zarray`, `.zattrs`, `.zgroup`, or `zarr.json` and then let ome-xml.js or zarrita.js "open" the array, parse (and maybe validate the JSON), and then give us details such as zarr version.

src/hooks/useZarrMetadata.ts
```diff
@@ -77,70 +145,43 @@ export default function useZarrMetadata() {
...
-                  );
-                  if (cancelRef.cancel) {
-                    return;
-                  }
-                  setThumbnailError('Error fetching OME-Zarr metadata');
+              if (attrs.node_type === 'array') {
+                checkZarrArray(imageUrl, 3, cancelRef);
+              } else if (attrs.node_type === 'group') {
+                if (
+                  attrs.attributes?.multiscales ||
```
mkitti, line 175: While Neuroglancer accepts `multiscales` outside of `ome` in zarr.json, I really think we should at least issue a warning here as we are clearly outside of the OME-NGFF specification when this occurs.

PR #164, opened 2025-09-12
bug: response body already consumed

allison-truhlar:
Clickup id: 86abrcxqh
@krokicki @neomorphic 

The body of the response was already getting consumed at the top of the `fetchFileInfo()` method, to provide info about the owner if there was a permissions error. Therefore, the error handling utility method `toHttpError()` couldn't consume it again when a non-permissions error was thrown to it. Now `fetchFileInfo()` uses the response body to construct a new `Error` directly, which is then displayed in the UI.


krokicki:
LGTM. I made a few updates to the way errors are presented to users, and added some defensive programming to the `toHttpError` method to deal with future bugs.

PR #165, opened 2025-09-14
Image content type detection

krokicki:
Clickup id: 86abtx3jk

Since the recently deployed content type detection was found to have some issues, @vaxenburg and I tried several other heuristics on a variety of image and segmentation data sets:

* **Unique Values Analysis (current)** - Fetched a center crop from the actual Zarr array data and calculated the percentage of unique pixel values. Low unique value ratios (<0.001) indicated segmentation data with large homogeneous regions.
  * Issues: 
    1) As chunk size increases, the latency to render the viewer links increases.
    2) It eventually runs out of memory when the chunks are large enough.

* **Compression Ratio Analysis** - Sampled multiple random chunks to calculate compression ratios by comparing compressed vs uncompressed chunk sizes. High compression ratios (>10) suggest segmentation data due to homogeneous regions compressing well.
  * Issues: 
    1) Fetching individual chunk sizes has too much latency to try more than 5-10, and that may not be enough. 
    2) Some volumes may be uncompressed.
    3) Some image volumes compress empty chunks instead of using fill value.

* **Thumbnail Unique Values Analysis** - Similar to current array-based analysis but operates on the thumbnail image. 
  * Issues: 
    1) Does not discriminate well between segmentation and natural images.

* **Sampled Thumbnail Unique Values Analysis** - Modified the method above to take multiple random tiny crops from the thumbnail and calculated the unique value ratio for each one, then take the maximum.
  * This had a 100% success rate on our test data, but it is stochastic.

* **Thumbnail Edge Content Analysis** - Analyzes edge content in thumbnails by shifting the image 1 pixel to the right,
  subtracting from the original, and calculating the ratio of pixels with differences. Low edge ratios (<0.05) indicate
  segmentation data, which typically has sharp boundaries between distinct labeled regions, while natural images have
  more continuous gradients and higher edge content.
  * This had a 100% success rate on our test data, and the largest discrimination between image classes.

@vaxenburg's edge detection approach was kept because it's fast, reliable (no randomness), and works well on our sample data.

I also added the detected layer type to the metadata table.

@StephanPreibisch @neomorphic @allison-truhlar

### DIFF
--- a/src/omezarr-helper.ts
+++ b/src/omezarr-helper.ts
@@ -449,14 +449,19 @@ async function getOmeZarrMetadata(dataUrl: string): Promise<Metadata> {
   const { arr, shapes, multiscale, omero, scales, zarr_version } =
     await omezarr.getMultiscaleWithArray(store, 0);
   log.debug(
-    'Zarr version: ', zarr_version, 
-    '\nArray: ', arr,
-    '\nShapes: ', shapes,
-    '\nMultiscale: ', multiscale,
-    '\nOmero: ', omero,
-    '\nScales: ', scales
+    'Zarr version: ',
+    zarr_version,
+    '\nArray: ',
+    arr,
+    '\nShapes: ',
+    shapes,
+    '\nMultiscale: ',
+    multiscale,
+    '\nOmero: ',
+    omero,
+    '\nScales: ',
+    scales
   );
-  
 
   const metadata: Metadata = {
     arr,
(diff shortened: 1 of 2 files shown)


src/omezarr-helper.ts
```diff
@@ -500,141 +507,106 @@ async function getOmeZarrThumbnail(
...
+
+        let nonZeroPixels = 0;
+        const totalPixels = img.width * img.height;
+
+        // Compare original and shifted images pixel by pixel
+        for (let i = 0; i < origData.data.length; i += 4) {
+          // Calculate difference for RGB channels (ignore alpha)
+          const rDiff = Math.abs(origData.data[i] - shiftData.data[i]);
+          const gDiff = Math.abs(origData.data[i + 1] - shiftData.data[i + 1]);
+          const bDiff = Math.abs(origData.data[i + 2] - shiftData.data[i + 2]);
```
vaxenburg, line 548: That's probably not a concern for an image a few hundred pixels in size, but just in case, I think, given how the shifting is done `ctx.drawImage(img, 1, 0)`, there will always be an artificial edge at x == 0 because the subtraction is done over the entire canvas, without excluding the x == 0 column?

vaxenburg: Right, and the extra edge probably changes the ratio by something like `1 / width`, so around 0.01 or 0.001, depending on the image. And the cutoff is 0.05 so it should be fine..

PR #166, opened 2025-09-15
feat: enable data link creation for all directories

allison-truhlar:
Clickup id: 86abny67z

This PR should be reviewed after PR #156 is merged. It builds off the automatic data link creation functionality.

This PR adds a toggle switch to the properties drawer to enable the user to make a proxied path for any directory that doesn't already have an external data link. Clicking the switch brings up the data link dialog, which either shows the "create" message or the "delete" message depending on whether a proxied path already exists for the directory.


krokicki (changes requested):
This works very nicely. Let's just add a "copy data URL" button in the properties panel.

PR #172, opened 2025-09-16
fix package lock

neomorphic:
Adds the missing optional dependencies, so that the package-lock will work on both macOS and linux. Also stop prettier from modifying the package-lock.json, so that it doesn't get marked as changed on every install.

- **fix: Updates the package lock to include optional deps.**
- **chore: make prettier check Ignore package-lock.json**
- **chore: commits un"prettier"ified package-lock.json**

@krokicki


krokicki (approved):
LGTM

PR #173, opened 2025-09-16
feat - add pagination to data links and tickets

allison-truhlar:
Clickup id: 86abmu11m

This PR enables pagination in the Tanstack table used for displaying data links and tickets, and adds buttons to navigate through the pages and a select element to change the number of entries to show per page.

In addition, with pagination now enabled, I removed the hardcoded cutoff of displaying 10 data links in the dashboard card, since this uses the same table component as the data links page.

@krokicki @neomorphic


krokicki:
Works great! A few thoughts:
* It would be nice if the column headers stayed visible at the top and the paginator stayed fixed at the bottom, so only the table’s content scrolls. That should make it much smoother to navigate.
* Adding a minimum height would help keep the layout usable on smaller screens.
* Showing the entire data links table on the front page feels a bit overwhelming and may cause performance issues in the future. What if we fetch only the 10 most recent links server side and display just the link itself here?

allison-truhlar:
@krokicki I refactored the Dashboard card to only show the links for the proxied paths. For now, I have not added the ability to fetch only the most recent 10 proxied paths from the API. As currently written, the ProxiedPath context is wrapped around the /browse path because both the data link dialog component (rendered downstream of the `<BrowseLayout/>` component) and the link page (downstream of the `<OtherPagesLayout/>`) require functions and data from the ProxiedPath context. I opened a ticket to refactor this, as I think it is a larger job outside the scope of this PR.

PR #181, opened 2025-09-26
fixes: central health server check; integration tests; package-lock version num

allison-truhlar:
Clickup id: 86abpt0c9

This PR was initiated to fix the Playwright integration tests. These were fixed (at least running locally in dev) by (1) ensuring the file-share-paths endpoint could include url params for the subpath and (2) adding specificity to the locator.

In the process of updating the Playwright tests, I ran into an issue related to the central server health check: if the central server url is not configured, the endpoints that call the central server return 500, which triggers the health check. Previously, the CENTRAL_SERVER_NOT_CONFIGURED error message would set the status back to health, triggering a page refresh, but then the endpoints that call the central server would return 500 again, and so on, resulting in endless refreshes. Now, the code looks for the response body and if it includes CENTRAL_SERVER_NOT_CONFIGURED, it does not trigger a health check, since this is the case where someone might be running Fileglancer locally.

### DIFF
--- a/fileglancer/handlers.py
+++ b/fileglancer/handlers.py
@@ -1,7 +1,6 @@
 import os
 import json
 import requests
-import re
 from datetime import datetime, timezone
 from abc import ABC
 from mimetypes import guess_type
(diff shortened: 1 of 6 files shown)


allison-truhlar:
Closing PR without merging - created a [new branch](https://github.com/JaneliaSciComp/fileglancer/tree/fix-playwright-tests) to address the Playwright errors

PR #192, opened 2025-10-03
feat: add user groups to api

allison-truhlar:
Clickup ids: 86aagg1rp and 86aafecg5
This PR is dependent on [PR #31 in fileglancer-central](https://github.com/JaneliaSciComp/fileglancer-central/pull/31).

This PR adds the user groups membership to the data returned by the /profile endpoint, and adds a "isFilteredByGroups" preference that defaults to True. In the client, the Zones are filtered by group membership when this preference is true. The user can toggle the preference on the Preferences page. There is a note at the bottom of the Zone browser indicating whether this preference is on or off with a link to the Preferences page for the user to modify the preference, if desired.

@krokicki

### DIFF
--- a/src/components/Preferences.tsx
+++ b/src/components/Preferences.tsx
@@ -150,7 +150,7 @@ export default function Preferences() {
               className="text-foreground"
               htmlFor="is_filtered_by_groups"
             >
-              Filter Zones by group membership
+              Display Zones for your groups only
             </Typography>
           </div>
 
--- a/src/components/ui/Sidebar/FavoritesBrowser.tsx
+++ b/src/components/ui/Sidebar/FavoritesBrowser.tsx
@@ -75,10 +75,10 @@ export default function FavoritesBrowser({
           displayFolders.length === 0 ? (
             <div className="px-4 py-6 text-center">
               <Typography className="text-sm text-gray-500">
-                No favorites match your filter.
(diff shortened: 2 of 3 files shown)


krokicki (changes requested):
The code looks correct but I ran into a fatal error that causes the application to stop:
Screenshot 2025-10-04 at 5 29 06 PM

This happens because the fsp is not defined for some favorites:
Screenshot 2025-10-04 at 5 30 45 PM

I'm guessing that this happens because I have favorites to folders outside of my groups.

Also, is there are any reason that the filtering must be server-side? What if we returned the groups as part of the user's profile (`/api/fileglancer/profile`) which is always requested? Then the filtering could be done quickly on the client side whenever the preference is changed, without additional requests.

krokicki:
I'm getting a different error now:

Screenshot 2025-10-07 at 11 08 15 AM

allison-truhlar:
@krokicki It should be fixed now - I'm not sure how but this [commit](https://github.com/JaneliaSciComp/fileglancer/pull/192/commits/62bb11846eee1ea74b4ff00d01ada193188f3979) I made to revert the file share paths handler back to no group filtering got reversed, so only FSPs in the user's group membership were being passed to the client by default again.

krokicki (approved):
Works great! I made a commit to tweak the wording a bit.

PR #196, opened 2025-10-06
fix: Restores breadcrumb to folders without access permission.

neomorphic:


neomorphic:
Please check this out and see if the UI makes sense to you when you view a closed directory

allison-truhlar:
I understand better what you were asking about now - I think it makes sense to use the blunt approach to hide those components for now. The cases that will be missed, I think, are where the user has read access but not write access at any level, and these will still be handled by toast errors when they try to use the "submit" buttons for any of those components.

neomorphic:
> The cases that will be missed, I think, are where the user has read access but not write access at any level, and these will still be handled by toast errors when they try to use the "submit" buttons for any of those components.

Correct, and we could check the file permissions when we load the FileOrFolder object and use the hasWrite/hasRead, but we would need to know the logged in users group at that point.

PR #220, opened 2025-10-17
fix: navigation input handling

allison-truhlar:
@krokicki @neomorphic 

Fixes a bug where a path with a dollar sign in it could not be navigated to from the navigation widget. This is because the useNavigationInput hook iterates through all the keys that start with fsp_ in the zonesAndFileSharePathsMap, and it used to check if the last segment of the key was found in the input path value. The last segment of the map key for a file share path with a dollar sign in the mac or the windows path will be "home", because that is how the name of the file share path is written (e.g., `groups_scicompsoft_home` corresponds to a Mac path of `smb://prfs.hhmi.org/scicompsoft$`). Since, the Mac and Windows path do not contain the string "home" in them, they were not flagged as a match. They were, however, flagged as a match for another file share path, in this case, `groups_scicompsoft_scicompsoft`. Since the mac path and the windows path for `groups_scicompsoft_scicompsoft` is exactly the same, minus the $ (e.g., `smb://prfs.hhmi.org/scicompsoft`), as the mac path and windows path for `groups_scicompsoft_home`, this segment was then stripped from the front of the input value and the rest of the input was interpreted as the sub-path, resulting in the $ getting separated from the mount path.

- Collect potential matches based on normalized paths, rather than trying to short-cut matching by comparing to the last portion of the file share path name
- Use the file share path mount path with the longest (most specific) match

### DIFF
--- a/src/hooks/useNavigationInput.ts
+++ b/src/hooks/useNavigationInput.ts
@@ -28,12 +28,12 @@ export default function useNavigationInput(initialValue: string = '') {
       // Trim white space and, if necessary, convert backslashes to forward slashes
       const normalizedInput = convertBackToForwardSlash(inputValue.trim());
 
-      // Collect all potential matches with their mount paths
-      const potentialMatches: Array<{
+      // Track best match
+      let bestMatch: {
         fspObject: FileSharePath;
         matchedPath: string;
         subpath: string;
-      }> = [];
+      } | null = null;
 
       const keys = Object.keys(zonesAndFileSharePathsMap);
       for (const key of keys) {
@@ -61,18 +61,22 @@ export default function useNavigationInput(initialValue: string = '') {
           }
 
           if (matchedPath) {
-            potentialMatches.push({ fspObject, matchedPath, subpath });
+            // The best match is the one with the longest matched path (most specific)
+            if (
+              !bestMatch ||
+              matchedPath.length > bestMatch.matchedPath.length
+            ) {
+              bestMatch = {
+                fspObject,
+                matchedPath,
+                subpath
+              };
+            }
           }
         }
       }
 
-      // If we have matches, use the one with the longest matched path (most specific)
-      if (potentialMatches.length > 0) {
-        potentialMatches.sort(
-          (a, b) => b.matchedPath.length - a.matchedPath.length
-        );
-        const bestMatch = potentialMatches[0];
-
+      if (bestMatch) {
         const browseLink = makeBrowseLink(
           bestMatch.fspObject.name,
           bestMatch.subpath


src/hooks/useNavigationInput.ts
```diff
@@ -21,41 +21,86 @@ export default function useNavigationInput(initialValue: string = '') {
...
+
+        // Convert mount paths to forward slashes for comparison
+        const normalizedLinuxPath = linuxPath
+          ? convertPathToPosixStyle(linuxPath)
+          : null;
+        const normalizedMacPath = macPath
+          ? convertPathToPosixStyle(macPath)
+          : null;
+        const normalizedWindowsPath = windowsPath
+          ? convertPathToPosixStyle(windowsPath)
```
krokicki: I would just move this normalization up to where windowsPath is created. The function could check for nulls and deal with them, to avoid this kind of boilerplate where every caller needs to wrap the method call in a ternary operator.

src/hooks/useNavigationInput.ts
```diff
@@ -21,41 +21,86 @@ export default function useNavigationInput(initialValue: string = '') {
...
+        // Collect all potential matches
+        if (
+          normalizedLinuxPath &&
+          normalizedInput.includes(normalizedLinuxPath)
+        ) {
+          matchedPath = normalizedLinuxPath;
+          subpath = normalizedInput.replace(normalizedLinuxPath, '').trim();
+        } else if (
+          normalizedMacPath &&
+          normalizedInput.includes(normalizedMacPath)
```
krokicki: Shouldn't this be startswith not includes?

krokicki (approved):
Looks good, I just made some suggestions below for code style that could help with readability.

neomorphic (approved):
I added a comment about extra loops, that probably doesn't matter in the grand scheme of things, but it would be more efficient.

src/hooks/useNavigationInput.ts
```diff
@@ -21,41 +21,86 @@ export default function useNavigationInput(initialValue: string = '') {
...
+        const linuxPath = fspObject.linux_path;
+        const macPath = fspObject.mac_path;
+        const windowsPath = fspObject.windows_path;
+
+        // Convert mount paths to forward slashes for comparison
+        const normalizedLinuxPath = linuxPath
+          ? convertPathToPosixStyle(linuxPath)
+          : null;
+        const normalizedMacPath = macPath
+          ? convertPathToPosixStyle(macPath)
```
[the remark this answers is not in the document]
allison-truhlar: Renamed the utility in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/95c93fbbe0e92e01bfd3ccdb39d51b9ca9f47109

src/hooks/useNavigationInput.ts
allison-truhlar: Added null checks in the utility function in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/95c93fbbe0e92e01bfd3ccdb39d51b9ca9f47109. Moved the normalization to where windowsPath is created in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/cd0c6c6a030d44fb2be021ec5c0c7efc64aa454a.

src/hooks/useNavigationInput.ts
```diff
@@ -21,41 +21,86 @@ export default function useNavigationInput(initialValue: string = '') {
...
-          const macPath = fspObject.mac_path;
-          const windowsPath = fspObject.windows_path;
+        const fspObject = zonesAndFileSharePathsMap[key] as FileSharePath;
+        const linuxPath = fspObject.linux_path;
+        const macPath = fspObject.mac_path;
+        const windowsPath = fspObject.windows_path;
+
+        // Convert mount paths to forward slashes for comparison
+        const normalizedLinuxPath = linuxPath
+          ? convertPathToPosixStyle(linuxPath)
```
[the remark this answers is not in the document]
allison-truhlar: Removed this in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/cd0c6c6a030d44fb2be021ec5c0c7efc64aa454a

src/hooks/useNavigationInput.ts
```diff
@@ -21,41 +21,86 @@ export default function useNavigationInput(initialValue: string = '') {
...
+          ? convertPathToPosixStyle(windowsPath)
+          : null;
+
+        let matchedPath: string | null = null;
+        let subpath = '';
+        // Check if the normalized input starts with any of the mount paths
+        // If a match is found, extract the subpath
+        // Collect all potential matches
+        if (
+          normalizedLinuxPath &&
```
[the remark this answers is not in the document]
allison-truhlar: Added this in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/cd0c6c6a030d44fb2be021ec5c0c7efc64aa454a

src/hooks/useNavigationInput.ts
allison-truhlar: Made this change in commit https://github.com/JaneliaSciComp/fileglancer/pull/220/commits/ba2be8544d82a990e55b180e59ef6e5a9fa7e301

PR #222, opened 2025-10-18
Frontend dir

krokicki:
The number of files in this repo's root directory keeps growing and makes it difficult to find what you're looking for. One way to deal with this would be to move the UI into a subdirectory, here named `frontend`. Making this work just requires modifying some of the config files and tasks to reference things inside the sub directory. I've sketched out a prototype in this PR. 

Are there any downsides to this approach, or anything I missed? Any alternatives we should consider?

@StephanPreibisch @neomorphic @allison-truhlar


allison-truhlar (approved):
I tried out the main pixi commands (install, build, dev-watch, dev-launch) and they worked fine for me with this change. I agree it will make it easier to find files in the root directory.

PR #235, opened 2025-10-30
refactor: migrate server state management to Tanstack Query

allison-truhlar:
This pull request introduces several improvements and refactors to the frontend codebase, focusing on migrating data fetching and state management to React/Tanstack Query.

**React Query Integration:**
* Added `@tanstack/react-query`, `@tanstack/react-query-devtools`, and `@tanstack/eslint-plugin-query` dependencies.
* Replaced usage of useEffects for data fetching with the queries.
* Updated frontend unit test utilities to wrap components in a `QueryClientProvider`.

**Miscellaneous:**
* Fixed Playwright tests by forcing them to run in parallel, have each test create a unique folder to generate test files in, and clearing the shared database at the end of each test.
* Added Playwright test report directory to `.prettierignore` to avoid formatting issues with generated reports.
* Fixed error on push: 'Can't find lefthook in PATH' by explicitly setting the lefthook location and command in `lefthook.yml`. This error resulted from the refactor that moved the frontend code, where lefthook is installed via `npm`.
* Added a frontend directory `CLAUDE.md` file to aid Claude Code in following the development patterns in this directory across collaborators.

**Note**: this PR now has several other PRs merged into it:
https://github.com/JaneliaSciComp/fileglancer/pull/236
https://github.com/JaneliaSciComp/fileglancer/pull/237
https://github.com/JaneliaSciComp/fileglancer/pull/238
https://github.com/JaneliaSciComp/fileglancer/pull/239

### DIFF
--- a/frontend/src/components/Browse.tsx
+++ b/frontend/src/components/Browse.tsx
@@ -70,19 +70,14 @@ export default function Browse() {
       className="flex flex-col h-full max-h-full w-full max-w-full"
       data-browse-container
       onPaste={async event => {
-        log.debug('React paste event fired!', event);
-
         // Check if any input, textarea, or contenteditable element is focused
         const activeElement = document.activeElement;
-        log.debug('Active element:', activeElement);
 
         const isTextInputFocused =
           activeElement &&
           (activeElement.tagName === 'INPUT' ||
             activeElement.tagName === 'TEXTAREA' ||
             activeElement.getAttribute('contenteditable') === 'true');
-
-        log.debug('Is text input focused:', isTextInputFocused);
 
         // Only handle paste if no text input is focused
         if (!isTextInputFocused) {
@@ -91,14 +86,12 @@ export default function Browse() {
 
           try {
             const clipboardText = await navigator.clipboard.readText();
-            log.debug('Clipboard text (API):', clipboardText);
             setPastedPath(clipboardText);
             setShowNavigationDialog(true);
           } catch (error) {
             log.debug('Clipboard API failed, using fallback:', error);
             // Fallback to event.clipboardData if clipboard API fails
             const clipboardText = event.clipboardData?.getData('text') || '';
-            log.debug('Clipboard text (fallback):', clipboardText);
             setPastedPath(clipboardText);
             setShowNavigationDialog(true);
           }
--- a/frontend/src/queries/zoneAndFileSharePathMapQuery.ts
+++ b/frontend/src/queries/zoneAndFileSharePathMapQuery.ts
@@ -95,7 +95,6 @@ export default function useZoneAndFileSharePathMapQuery(): UseQueryResult<
 
   const makeSortedZonesAndFspsMap = (data: { paths: FileSharePath[] }) => {
     const zonesAndFileSharePathsMap = createZonesAndFileSharePathsMap(data);
-    console.log('Zones and File Share Paths Map:', zonesAndFileSharePathsMap);
     return alphabetizeZonesAndFsps(zonesAndFileSharePathsMap);
   };
 


frontend/src/components/Help.tsx
```diff
@@ -29,14 +29,14 @@ export default function Help() {
     {
       icon: TbBrandGithub,
       title: 'Release Notes',
-      description: `What's new and improved in Fileglancer version ${versionNo}`,
-      url: `https://github.com/JaneliaSciComp/fileglancer/releases/tag/${versionNo}`
+      description: `What's new and improved in Fileglancer version ${versionQuery.data?.version}`,
+      url: `https://github.com/JaneliaSciComp/fileglancer/releases/tag/${versionQuery.data?.version}`
```
neomorphic: There is no loading check to make sure the data has loaded. This could result in a broken link to https://github.com/JaneliaSciComp/fileglancer/releases/tag/ if the user is on a slow connection.

frontend/src/components/ui/PropertiesDrawer/PropertiesDrawer.tsx
```diff
@@ -88,9 +88,12 @@ export default function PropertiesDrawer({
...
   const { fileBrowserState } = useFileBrowserContext();
   const { pathPreference, areDataLinksAutomatic } = usePreferencesContext();
-  const { ticket } = useTicketContext();
-  const { proxiedPath, dataUrl } = useProxiedPathContext();
+  const { ticketByPathQuery } = useTicketContext();
+  const { proxiedPathByFspAndPathQuery } = useProxiedPathContext();
   const { externalDataUrl } = useExternalBucketContext();
+
+  const ticket = ticketByPathQuery.data;
+  const proxiedPath = proxiedPathByFspAndPathQuery.data;
```
neomorphic: How are you making sure that the .data is loaded for these. I would expect to see this use the isPending state to decide if the data has been loaded. I didn't notice an isPending check in the TicketsContext that would prevent this component from loading and passing the incorrect ticket value to TicketDetails.

frontend/src/queries/fileQueries.ts
```diff
@@ -0,0 +1,463 @@
...
+    } else if (response.status === 404) {
+      throw new Error('Folder not found');
+    } else {
+      throw new Error(
+        body.error ? body.error : `Unknown error (${response.status})`
+      );
+    }
+  };
+
+  const transformData = React.useCallback(
```
krokicki: Does this need to be a Callback? Couldn't it be a regular function?

frontend/src/contexts/PreferencesContext.tsx
```diff
@@ -29,40 +34,49 @@ export type FolderPreference = {
...
+    localPathPreference: PreferencesContextType['pathPreference']
+  ) => Promise<Result<void>>;
+  handleUpdateLayout: (layout: string) => Promise<void>;
+  setLayoutWithPropertiesOpen: () => Promise<Result<void>>;
+  toggleHideDotFiles: () => Promise<Result<void>>;
+  toggleAutomaticDataLinks: () => Promise<Result<void>>;
+  toggleDisableNeuroglancerStateGeneration: () => Promise<Result<void>>;
+  toggleDisableHeuristicalLayerTypeDetection: () => Promise<Result<void>>;
+  toggleUseLegacyMultichannelApproach: () => Promise<Result<void>>;
+  toggleFilterByGroups: () => Promise<Result<void>>;
```
krokicki: This is so much more readable than before!

frontend/src/queries/preferencesQueries.ts
```diff
@@ -0,0 +1,492 @@
...
+}
+
+/**
+ * Mutation hook for updating recently viewed folders with optimistic updates
+ *
+ * @example
+ * const mutation = useUpdateRecentlyViewedFoldersMutation();
+ * mutation.mutate(updatedFolders);
+ */
+export function useUpdateRecentlyViewedFoldersMutation(): UseMutationResult<
```
krokicki: It looks like this and `useUpdateFavoritesMutation` are basically the same code duplicated. Can they be combined?

frontend/src/hooks/useNewFolderDialog.ts
neomorphic: This change removes all the code that was handling the error states and returning a Result object from handleNewFolderSubmit. This means that you had to reintroduce a try /catch block in the [NewFolderButton.tsx](https://github.com/JaneliaSciComp/fileglancer/pull/235/files#diff-b5177406bcd35cf2b8dea20eb8268322f5d04367ed6d83904c4798c06452cc01) I thought the original refactoring was supposed to get rid of the try blocks in components. Is it possible to move that logic into this hook?

frontend/src/queries/fileQueries.ts
```diff
@@ -0,0 +1,463 @@
...
+    },
+    [fspName, zonesAndFspQuery.data]
+  );
+
+  const query = useQuery<FileBrowserResponse, Error, FileQueryData>({
+    queryKey: fileQueryKeys.filePath(fspName || '', folderName),
+    queryFn: fetchFileInfo,
+    select: (data: FileBrowserResponse) => transformData(data),
+    enabled: !!fspName && !!zonesAndFspQuery.data,
+    staleTime: 5 * 60 * 1000 // 5 minutes - file listings don't change that often
```
krokicki: This is an interesting choice and maybe a good one (since you can still get updates with the Refresh button), but  let's discuss this some more at the next meeting. I'm also interested in all the other staleTime settings, and if the Refresh workaround works in all cases.

frontend/src/components/ui/BrowsePage/Toolbar.tsx
```diff
@@ -132,20 +138,18 @@ export default function Toolbar({
                 const isViewingFile =
```
neomorphic: Is it possible that all the logic in the refresh onClick could be moved into the  useFileBrowserContext and imported as a function from there? Then this component wouldn't need to import useQueryClient.

neomorphic (changes requested):
I ran the code and it looks like it is working. I left a number of comments in places that might need looking at. I was not able to look at all the test, but the ones I did look at seem okay.

frontend/src/App.tsx
```diff
@@ -71,45 +71,62 @@ const AppComponent = () => {
...
           <Route
             element={
-              <RequireAuth>
-                <BrowsePageLayout />
-              </RequireAuth>
+              <ErrorBoundary FallbackComponent={ErrorFallback}>
+                <RequireAuth>
+                  <BrowsePageLayout />
+                </RequireAuth>
+              </ErrorBoundary>
```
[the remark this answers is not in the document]
allison-truhlar: This was added as a result of asking Claude to review all my Tanstack Query changes and suggest improvements - Claude suggested adding a global `throwOnError: true` setting to the Tanstack Query client configuration, which would then required route-specific error boundaries so that you could still navigate to other routes. However, this turned out to be a very bad suggestion since our app has some expected errors, like the 403 errors when navigating to a folder/file you don't have permission to view. I was handling this error in the query, but with the throwOnError, it was causing the full page error boundary to show instead - I will open a separate PR to fix this since Konrad put it in as a ClickUp ticket. Here, I will remove the error boundaries around each route.

frontend/src/components/ui/Dialogs/TextWithFilePath.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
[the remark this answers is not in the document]
allison-truhlar: I change the React type to use React 18. React 18 is still required by the Material Tailwind component library. PR: https://github.com/JaneliaSciComp/fileglancer/pull/238

frontend/ui-tests/tests/load-zarr-files.spec.ts
```diff
@@ -7,30 +7,48 @@ test.describe('Zarr File Type Representation', () => {
...
+      await page.mouse.move(0, 0);
     }
   );
 
   test('Zarr V3 with no OME metadata should show only neuroglancer', async ({
     fileglancerPage: page
   }) => {
-    await page.getByText(ZARR_TEST_FILE_INFO.v3_non_ome.dirname).click();
+    // Move mouse away to avoid tooltip interference
+    await page.mouse.move(0, 0);
```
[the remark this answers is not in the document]
allison-truhlar: Sometimes the tooltip would block the visibility of the link I wanted Playwright to click on. Adding these commands to move the mouse away and then try to click on the link ensured this didn't happen.

frontend/ui-tests/utils/db-cleanup.ts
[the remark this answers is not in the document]
allison-truhlar: Yes - the file_share_paths table, which is generated when the server starts up from the `FGC_FILE_SHARE_MOUNTS` environmental variable in playwright.config.js (from `config.yaml` in the normal app start up).

frontend/src/components/ui/BrowsePage/FileBrowser.tsx
```diff
@@ -53,12 +53,13 @@ export default function FileBrowser({
     layerType
   } = useZarrMetadata();
 
+  const currentFileOrFolder = fileQuery.data?.currentFileOrFolder;
+  const isLoading = fileQuery.isPending;
+  const error = fileQuery.error;
```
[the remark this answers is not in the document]
allison-truhlar: I think I reordered the checks correctly in this commit: https://github.com/JaneliaSciComp/fileglancer/pull/235/commits/f6e91ed8e1de23a4d939886d1196eb995978cc91

frontend/src/components/ui/BrowsePage/Toolbar.tsx
allison-truhlar: I moved the logic out into a hook here: https://github.com/JaneliaSciComp/fileglancer/pull/235/commits/f79975ba170a03bcdf524b4da706305ed97f650f

frontend/src/queries/fileQueries.ts
allison-truhlar: Yes - fixed this here: https://github.com/JaneliaSciComp/fileglancer/pull/235/commits/06010939c755c6cb8c00cb56a17a3262e0f2a002

frontend/src/contexts/ExternalBucketContext.tsx
```diff
@@ -33,110 +31,60 @@ export const ExternalBucketProvider = ({
...
+    !fileQuery.isPending && !!fileBrowserState.uiFileSharePath
   );
-  const { fileBrowserState } = useFileBrowserContext();
 
-  const updateExternalBucket = React.useCallback(
-    (bucket: ExternalBucket | null) => {
-      setExternalBucket(bucket);
+  // Compute external data URL based on query data
+  const externalDataUrl = React.useMemo(() => {
+    const bucket = externalBucketQuery.data;
```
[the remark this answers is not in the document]
allison-truhlar: Fixed this in commits https://github.com/JaneliaSciComp/fileglancer/pull/235/commits/47cf359b8fdda4061bdf019434d2c9d65eaeea48 and https://github.com/JaneliaSciComp/fileglancer/pull/235/commits/6076ddb59c217f993154e978b2c300f67197e086

PR #238, opened 2025-11-05
fix: React setup

allison-truhlar:
@krokicki @neomorphic 

This PR addresses @neomorphic's [comment](https://github.com/JaneliaSciComp/fileglancer/pull/235#discussion_r2481879269) regarding the two different versions that were being used for React vs React types. Now version 18 is used for both. Updating to version 19 is not currently possible because Material Tailwind still requires React 18.

It also updates the code to: 
1. Follow modern React best practice of not importing the full React namespace, only the needed types and methods,
2. Standardize the return type of React function components by not providing a return type (allowing inference), unless absolutely necessary, following the recommendation [here](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript).
3. Standardize type imports to be done on a separate line from method imports from the same module.

These practices are added to `frontend/CLAUDE.md` to promote standardization across collaborators using Claude Code.

### DIFF
--- a/frontend/src/components/ui/BrowsePage/Toolbar.tsx
+++ b/frontend/src/components/ui/BrowsePage/Toolbar.tsx
@@ -1,5 +1,4 @@
 import { useMemo } from 'react';
-import type { MouseEvent } from 'react';
 import toast from 'react-hot-toast';
 import { Link } from 'react-router';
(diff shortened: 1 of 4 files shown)


neomorphic (approved):
LGTM

PR #241, opened 2025-11-07
fix: url encoding for data tools

allison-truhlar:
Clickup id: 86ad24hev
@krokicki @neomorphic 

This pull request focuses on improving URL encoding and decoding consistency across both the backend and frontend, which addresses interoperability issues with external tools (like Vol-E) and ensures that API requests handle special characters correctly. The changes also include a minor simplification to a mutation function in the frontend.

**Path for testing** (the data link created from this path used to not work due to improper encoding of the % signs. This path also had problems being accessed by Vol-E): `/groups/liconn/liconn/data_internal/FlyLICONN_zarr_converted/2November2025/20251019_FlyID11_2nd_0.2%-0.2%GMA_2%AA_40XW.zarr`

**Backend changes for URL decoding:**

* Updated `_get_file_proxy_client` in `fileglancer/app.py` to handle both properly encoded and Vol-E-style unencoded sharing names, improving compatibility with external tools that send literal `%` characters in URLs. 

**Frontend changes for URL encoding:**

* Updated all API calls in `frontend/src/queries/externalBucketQueries.ts`, `proxiedPathQueries.ts`, and `ticketsQueries.ts` to use `encodeURIComponent` for `fspName` and `path` query parameters, ensuring safe transmission of special characters. 
* Modified `useZarrMetadata.ts` so that URLs passed to external tools that use URL query parameters (validator, Vol-E, Avivator) are encoded with `encodeURIComponent`, preventing issues with special characters in dataset paths. 

**Misc:**

* Simplified the mutation function in `useCreateTicketMutation` by removing the unused `signal` parameter.

### DIFF
--- a/frontend/src/utils/index.ts
+++ b/frontend/src/utils/index.ts
@@ -232,137 +232,70 @@ function makeMapKey(type: 'zone' | 'fsp' | 'folder', name: string): string {
 }
 
 /**
- * Constructs an API URL with proper encoding for path segments and query parameters.
- * This is the recommended way to build URLs for use with sendFetchRequest.
+ * Constructs a properly encoded URL from base URL, optional path, and optional query parameters.
  *
- * @param basePath - The base API path (e.g., '/api/files/')
- * @param pathSegments - Optional path segments to append (e.g., fspName)
- * @param queryParams - Optional query parameters as key-value pairs
- * @returns A properly encoded URL string ready for sendFetchRequest
+ * Two overloads:
+ * 1. URLs with a single path segment and/or query parameters
+ * 2. URLs with multi-segment path strings (S3-style)
+ *
+ * @param baseUrl - The base URL (e.g., '/api/files/' or 'https://viewer.example.com/')
+ * @param singlePathSegment - Single path segment, encoded with encodeURIComponent (Overload 1)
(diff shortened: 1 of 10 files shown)


frontend/src/utils/index.ts
```diff
@@ -234,12 +231,155 @@ function makeMapKey(type: 'zone' | 'fsp' | 'folder', name: string): string {
...
+ * // Path with query parameters
+ * buildApiUrl('/api/files/', ['myFSP'], { subpath: 'folder/file.txt' })
+ * // Returns: '/api/files/myFSP?subpath=folder%2Ffile.txt'
+ *
+ * @example
+ * // Multiple query parameters
+ * buildApiUrl('/api/ticket', [], { fsp_name: 'myFSP', path: 'folder/file.txt' })
+ * // Returns: '/api/ticket?fsp_name=myFSP&path=folder%2Ffile.txt'
+ */
+function buildApiUrl(
```
neomorphic: buildApiUrl, buildExternalUrlWithQuery & buildExternalUrlWithPath all look very similar. Could these be combined into a single function with different arguments? It would be less of a mental load when trying to remember which function name to call.

neomorphic: This looks much better to me. I would go with it.

PR #242, opened 2025-11-11
fix: clicking on filtered data link table does not reset state

allison-truhlar:
Clickup ids: 86acthka5, 86acwmgfg

This PR changes the table filtering UI to a single search input that searches across all the table columns. It fixes the problem of the filter state clearing when the user clicked outside the search input - the filter state only clears (and subsequently resets the table) if you explicitly click the clear button. The PR also adds an end-to-end test to verify these behaviors.

### DIFF
--- a/frontend/ui-tests/tests/file-operations.spec.ts
+++ b/frontend/ui-tests/tests/file-operations.spec.ts
@@ -68,6 +68,6 @@ test.describe('File Operations', () => {
     await expect(page.getByRole('dialog')).not.toBeVisible();
 
     // Verify the new folder appears in the file list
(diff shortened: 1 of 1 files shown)


neomorphic (commented):
The fix works, but I question if it is the correct UI? Should clicking outside of the filter pop up clear the filter or should it simply close the filter pop up and leave the filter in place? I think leaving the filter in place is what I would expect from the UI. Maybe add a small red dot to the search icon to indicate that a filter has been applied. Then, if I wish to remove the filter I can click on the icon and clear the text from the pop over.

PR #251, opened 2025-11-19
feat: make table values copy-able

allison-truhlar:
Clickup id: 86aceaxy3

This PR makes the values in the data links and task tables copy-able in two ways:
1. Click on the text in a cell (with the exception of the linked paths) to automatically select the entire text. Then use Crtl + C to copy.
2. Right click in any cell and select "copy" from the context menu.

In the process of making this change, I refactored the copy tooltip logic to a hook and made a `CopyTooltip` reusable component (I had originally thought I would go the "click to copy" route, and use the tooltip to indicate the text had been copied). I also refactored the context menu from the `FileBrowser` to make it reusable for the tables.

@krokicki

### DIFF
--- a/frontend/src/components/ui/Table/linksColumns.tsx
+++ b/frontend/src/components/ui/Table/linksColumns.tsx
@@ -68,19 +68,6 @@ function PathCell({
     data: { value: string }
   ) => void;
 }) {
-  const { pathPreference } = usePreferencesContext();
-  const { zonesAndFspQuery } = useZoneAndFspMapContext();
-
-  const pathFsp = zonesAndFspQuery.isSuccess
-    ? (zonesAndFspQuery.data[makeMapKey('fsp', item.fsp_name)] as FileSharePath)
-    : undefined;
-
-  const displayPath = getPreferredPathForDisplay(
-    pathPreference,
-    pathFsp,
-    item.path
-  );
-
   const browseLink = makeBrowseLink(item.fsp_name, item.path);
 
   return (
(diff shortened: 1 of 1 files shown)


frontend/src/components/ui/Table/TableCard.tsx
```diff
@@ -296,6 +318,33 @@ function Table<TData>({
   const timerRef = useRef<NodeJS.Timeout | null>(null);
   const inputRef = useRef<HTMLInputElement>(null);
 
+  // Context menu state
+  const {
+    contextMenuCoords,
```
neomorphic: Why are you importing the contextMenuCoords here, only to pass them on to the ContextMenu component. Couldn't you have the ``` const {contextMenuCoords} = useContextMenu<CellContextMenuData>();``` call in the ContextMenu component? What am I missing?

allison-truhlar: I've found when working with hooks if there is state that is operated on by functions also exposed by the hook, the behavior gets weird if you don't create the state and function together in one component and pass them down together, if needed. So specifically for the `showContextMenu` state, if I create fresh state in the `ContextMenu` by calling the hook again, the close function is not operating on the `showContextMenu` state in `TableCard`, so the context menu stays visible.

allison-truhlar: https://react.dev/learn/reusing-logic-with-custom-hooks#custom-hooks-let-you-share-stateful-logic-not-state-itself

PR #253, opened 2025-11-20
style: zarr metadata table display; delete btn colors; right handle padding

allison-truhlar:
Clickup ids: 86adbzmdy; 86adc0g69

This PR fixes three style issues:

1. The `ZarrPreview` expands to the full width of the file table to prevent cutting off the metadata tables. The metadata tables are displayed side-by-side to save vertical space when the main panel is sufficiently wide.
2. Creates a reusable `DeleteBtn` for use in the `Delete` and `DataLink` dialogs. The button does not turn to white while the delete is pending. Also fixed the pending state for deleting data links, to include the time required to "refresh" the data links after deleting.
3. The padding on the right panel handle is slightly increased to make up for the apparently wider left panel handle. This appearance is caused by the grey inner border of the `SidePanel` component used in the left panel.

@krokicki

### DIFF
--- a/frontend/ui-tests/utils/navigation.ts
+++ b/frontend/ui-tests/utils/navigation.ts
@@ -5,10 +5,10 @@ const navigateToScratchFsp = async (page: Page) => {
   const localZone = page
     .getByLabel('List of file share paths')
     .getByRole('button', { name: 'Local' });
-  await localZone.click();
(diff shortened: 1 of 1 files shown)


neomorphic (approved):
Looks good to me. All the UI changes work as advertised and I was not able to break them.

PR #254, opened 2025-11-20
fix: prevent page access with expired session

neomorphic:
Clickup ID: [86add3995](https://app.clickup.com/t/86add3995)  
Changes session expiry handling to update auth state instead of hard redirect:

- Exports queryClient from main.tsx to make it accessible for auth state updates
- Updates checkSessionValidity() to use /api/auth/status endpoint instead of /api/profile
  for more reliable session validation
- Parses authenticated boolean from response JSON rather than relying on HTTP status
- Handles both 403 and 401 status codes as potential session expiry indicators
- Replaces window.location.href redirect with queryClient.setQueryData() to update
  auth status in React Query cache, allowing the app to gracefully handle expired
  sessions through the existing auth flow

@krokicki

### DIFF
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -190,7 +190,7 @@ You can also run these in UI debug mode using:
 pixi run test-ui -- --ui --debug
 ```
 
-If you are unable to use the UI mode, record a trace for inspecting in the [Playwright trace viewer](https://trace.playwright.dev):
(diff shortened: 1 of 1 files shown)


neomorphic:
@krokicki @allison-truhlar 
The easiest way to test this is to:

1. open the Application tab of the developers tools. 
2. Find the Cookie called fg_session.
3.  modify the hashed value by adding some random characters
4. navigate to another page on the site.

This should force the user to the login page.

allison-truhlar (approved):
Works for me!

PR #259, opened 2025-12-02
fix: UI error when no proxy URL is configured

allison-truhlar:
Clickup id: 86adg5gqz
@krokicki 

This fix addresses a bug where, if you didn't provide an `external_proxy_url` in the `config.yaml`, when you create a data link, you would get a success message and (if using the data link toggle in the properties drawer) the data link toggle would indicate that a link was created. However, if you then went to click on a data link tool to view the data, you then get a toast error message stating "URL not available". The issue was that the data link was never made, but this error caused by the missing `external_proxy_url` was not raised from the server to the client. 

This PR makes `external_proxy_url` required in the `Settings` model in `settings.py` - the app will not start if this setting is not provided. It also ensures that others errors from the proxied path methods on the server will be propertly displayed in in the toast message when using the data link dialog, in the properties panel, and in the data link table. It also adds the `external_proxy_url` environmental variable to the Playwright and backend testing setups.

### DIFF
--- a/fileglancer/settings.py
+++ b/fileglancer/settings.py
@@ -1,7 +1,8 @@
 from typing import List, Optional
 from functools import cache
+import sys
 
-from pydantic import HttpUrl, model_validator
+from pydantic import HttpUrl, ValidationError, field_validator, model_validator
 from pydantic_settings import (
     BaseSettings,
     PydanticBaseSettingsSource,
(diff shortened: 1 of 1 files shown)


fileglancer/settings.py
```diff
@@ -43,7 +43,7 @@ class Settings(BaseSettings):
     
     # The external URL of the proxy server for accessing proxied paths.
     # Maps to the /files/ end points of the fileglancer-central app.
-    external_proxy_url: Optional[HttpUrl] = None
+    external_proxy_url: HttpUrl = 'http://localhost:7878/files/'
```
neomorphic: As far as I can tell, this makes the external_proxy_url "optional" by setting a default. If I leave it out, the server will start up and use the default value that is set here. Is that the functionality that we want? I was thinking that the server would require that this value be populated in the config.yaml or set as FGC_EXTERNAL_PROXY_URL. If either of those are missing, then the server should print a warning that the configuration is missing and shutdown.

PR #263, opened 2025-12-10
fix: show when auto-detect content is disabled in zarr metadata table

allison-truhlar:
Clickup id: 86adp8k8t
@krokicki 

This PR restores the [original logic](https://github.com/JaneliaSciComp/fileglancer/commit/2e056a0168e44aed2ad9fca3e261203571035908#diff-b5db4aa20fe5ea708e6a4feb70bc4917def6c25cebaef2f37d91cf0ef96ab900) for the determining the content type in `useZarrMetadata` and shows in the metadata table when the `disableHeuristicalLayerTypeDetection` is `true`. It also adds a tooltip in this case, indicating that the user can go to the preferences page to change this, with a link. It also adds component tests to make sure the metadata table shows the correct information based on the value of `disableHeuristicalLayerTypeDetection`.


neomorphic (approved):
Looks good to me and the link to the preferences worked well.

PR #265, opened 2025-12-12
feat: file selection dialog

allison-truhlar:
Clickup id: 86acw956g

This PR adds a file selection "Browse" button that, when clicked, displays a mini file browser in a dialog. This is integrated into the "Convert File" dialog, allowing the user to navigate to select a destination path. The initial file system location displayed in the dialog is the current location in the main file browser. 

In the course of adding this feature, several components were refactored to pull out common logic that can be shared between the mini file browser and the main file browser. Specifically, the rendering logic for a breadcrumb segment, the rendering logic for three of the file browser table columns (typeColumn, lastModifiedColumn, and sizeColumn - the name column is different for each because the main file browser does not need to render file share paths and zones, and the actions column is not needed in the mini file browser), and the logic to filter out zones that the user is not a member of.

@krokicki

### DIFF
--- a/frontend/src/__tests__/componentTests/ConvertFile.test.tsx
+++ b/frontend/src/__tests__/componentTests/ConvertFile.test.tsx
@@ -66,6 +66,10 @@ describe('Convert File dialog', () => {
       screen.getByPlaceholderText('/path/to/destination/folder/'),
       '/test'
     );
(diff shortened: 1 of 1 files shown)


neomorphic (approved):
Looks good to me.

PR #266, opened 2025-12-12
feat: data link code snippets dialog

allison-truhlar:
Clickup ids: 86acuxatq, 86accr60r, 86aeq5001, 86accr5yg

This PR adds a data link usage dialog with copy-able code snippets and instructions for opening data links in external tools. The dialog is organized into tabs by tool/language (Fiji, Java, Napari, Python, and VVDViewer) and is context-aware — showing different content for zarr, N5, and directory data types. This determination is owned by the dialog - it uses the `detectZarrVersions` and `detectN5` methods from `/queries/fileQueries.ts`. In the dialog header, viewable no matter which tab is selected, is a copy data link box. The dialog is accessible from the zarr/N5 preview panels, the properties drawer, and the data links table.                                                                                                                                                              
                                                                                                                                                                                                                                                                                  
This PR also redesigns the data tool link icons (Neuroglancer, Vol-E, Avivator, etc.) as labeled circular buttons with hover effects, replacing the previous ButtonGroup layout. It adds the "More ways to open" button, displayed as a three-dot "More..." icon in the Zarr/N5 preview panels and a button with text in the properties panel, that opens the data link usage dialog. 

Finally, this PR: 
- adds two reusable dialog button components (DialogIconBtn, DialogTextBtn)
- fixes data links in the properties panel for directories inside `data_external` folders. Previously, if you clicked a sub-directory row of the current directory in the file browser, the properties panel did not update to show the link for that sub-directory. This is already fixed on main for data links that are not S3 proxy links (external data links).

@krokicki @neomorphic @cgoina

### DIFF
--- a/frontend/src/components/ui/Dialogs/dataLinkUsage/tabsContent/javaTab/JavaTab.tsx
+++ b/frontend/src/components/ui/Dialogs/dataLinkUsage/tabsContent/javaTab/JavaTab.tsx
@@ -28,7 +28,7 @@ export default function JavaTab({
   const prerequisites = [
     {
       label: 'Java Development Kit (JDK)',
-      href: 'https://jdk.java.net/'
+      href: 'https://www.azul.com/downloads/?version=java-21-lts&package=jdk-fx#zulu'
     },
     {
       label: 'Set Java environmental variables',


frontend/src/components/ui/BrowsePage/DataToolLinks.tsx
```diff
@@ -1,155 +1,187 @@
...
-          >
-            <Link
-              onClick={async e => {
-                e.preventDefault();
-                await onToolClick('neuroglancer');
-              }}
-              rel="noopener noreferrer"
-              target="_blank"
-              to={urls.neuroglancer}
+          <div className="flex flex-col items-center w-16">
```
neomorphic: There is a div for each of the viewers in this file. Maybe these could be extracted into their own components to reduce the amount of code that is in this return block.

PR #269, opened 2025-12-15
fix: refresh zarr queries with file browser refresh

allison-truhlar:
Clickup id: 86adxf9gw

This PR invalidates the Zarr queries when the 'Refresh' button is clicked in the toolbar, triggering a refetch of the Zarr metadata and thumbnail if the file browser is currently on a Zarr directory. It also adds a Playwright test to check this functionality.

@krokicki

### DIFF
--- a/frontend/src/components/ui/BrowsePage/Toolbar.tsx
+++ b/frontend/src/components/ui/BrowsePage/Toolbar.tsx
@@ -145,6 +145,7 @@ export default function Toolbar({
         <ButtonGroup className="gap-1">
           {/* Show/hide favorites and zone browser sidebar */}
           <FgTooltip
+            as={IconButton}
             icon={showSidebar ? GoSidebarExpand : GoSidebarCollapse}
             label={
               showSidebar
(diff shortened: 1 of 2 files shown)


frontend/src/components/ui/BrowsePage/Toolbar.tsx
```diff
@@ -173,6 +173,7 @@ export default function Toolbar({
           {/* Refresh browser contents */}
           {currentFileSharePath ? (
             <FgTooltip
+              as={IconButton}
```
neomorphic: Was this suggested by a lint rule? I presume it is to make sure that this clickable element is a button instead of a div to aid with accessibility. If so, it looks like a number of the other items in the list are also divs and should be updated in a similar manner.

allison-truhlar: No, I added this to make the element easier to locate for Playwright testing. But you're right - I'll change the others to buttons as well.

PR #271, opened 2025-12-16
fix: remove unneeded e.currentTarget.blur() calls

allison-truhlar:
Clickup id: 86adx9mee

This PR removes the `e.currentTarget.blur()` calls from the end of each `onClick` method for the Toolbar buttons. This has the benefit of eliminating the error printed to the console related to the target being null for the view dot file button. This removal did not change the visual appearance of the buttons/tooltips, and is better for accessibility by not breaking the expected tab behavior. Further, one button (add new folder), did not have `e.currentTarget.blur()`, so this makes the button logic more consistent.

@krokicki


neomorphic (approved):
LGTM!

PR #280, opened 2026-01-05
Pretty print JSON in the file viewer

krokicki:
A little feature that automatically formats JSON (pretty print). This was requested by @mkitti and can be especially useful for N5 `attributes.json` where everything is on one line. It can be toggled off in case it causes any issues for anyone.

@StephanPreibisch @mkitti @neomorphic

### DIFF
--- a/frontend/src/components/ui/BrowsePage/FileViewer.tsx
+++ b/frontend/src/components/ui/BrowsePage/FileViewer.tsx
@@ -135,11 +135,22 @@ export default function FileViewer({ file }: FileViewerProps) {
       }
     }
 
+    // Get the theme's code styles and merge with padding bottom for scrollbar
+    const theme = isDarkMode ? materialDark : coy;
+    const themeCodeStyles = theme['code[class*="language-"]'] || {};
+    const mergedCodeTagProps = {
+      style: {
+        ...themeCodeStyles,
+        paddingBottom: '2em'
+      }
+    };
+
     return (
       <SyntaxHighlighter
+        codeTagProps={mergedCodeTagProps}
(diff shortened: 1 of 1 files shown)


mkitti:
Note that this also applies to Zarr as well. TensorStore writes minified JSON to zarr.json by default. For example see [`/groups/tavakoli/tavakolilab/data_external/Cortex_Data_unpublished/ExPID124_image.zarr/s0/zarr.json`](https://s3.janelia.org/tavakoli-lab/Cortex_Data_unpublished/ExPID124_image.zarr/s0/zarr.json)

allison-truhlar (approved):
LGTM! I added one commit that I think achieves what you were trying to do by adding bottom padding to the `customStyles` prop. This added padding to the bottom of the viewer, but the scrollbar still rendered over the last line of text. Adding the bottom padding to `codeTagProps` does move the scrollbar beneath the last line of text. Note that this prop completely overwrites the default styles of the code tag, which is why I needed to get the theme styles and then add the bottom padding to those existing styles.

krokicki:
Thanks for fixing that @allison-truhlar! I didn't test with a narrow browser width so I didn't catch that, I was just trying to add some padding so the file text wouldn't run to the very edge of the window. 

It's functional at narrow width, but looks a little strange:
Screenshot 2026-01-05 at 6 28 24 PM

It's probably not worth tweaking more right now (I want to merge this so that I can use it for developing another PR), but maybe it's something we can revisit later.

mkitti:
What happens if the JSON is malformed? It seems to silently fail. That's not the worse outcome but it could be confusing.

PR #285, opened 2026-01-05
fix(frontend): handle non-JSON error responses in file fetches

### LINKED ISSUE #283: fileglancer attempts to parse the returned response contents as JSON despite a response code of 500.
Below, fileglancer is attempting to parse the text "Internal Server Error" as JSON:

Image

mkitti:
Fix #283 so that error responses do not result in JSON parsing errors of 
that error response being displayed.

Before:
Image

After:

image

Assisted by Gemini 3 Pro (Low) via Antigravity.

### DIFF
--- a/fileglancer/app.py
+++ b/fileglancer/app.py
@@ -289,6 +289,12 @@ def create_app(settings):
         return JSONResponse({"error": f"Permission denied: {error_msg}"}, status_code=403)
 
 
+    @app.exception_handler(Exception)
+    async def general_exception_handler(request, exc):
+        logger.exception(f"Unhandled exception: {exc}")
+        return JSONResponse({"error": f"{type(exc).__name__}: {str(exc)}"}, status_code=500)
+
+
     @app.get('/robots.txt', response_class=PlainTextResponse, include_in_schema=False)
     def robots():
         return """User-agent: *\nDisallow: /"""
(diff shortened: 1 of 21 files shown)


krokicki:
I think it's useful to do some defensive programming here, but we should also try to understand why this response is being generated, and change it so that it generates JSON. 

The Starlette exception handlers should be doing this, and it seems that is not working somehow.
```
    @app.exception_handler(StarletteHTTPException)
    async def http_exception_handler(request, exc):
        return JSONResponse({"error":str(exc.detail)}, status_code=exc.status_code)


    @app.exception_handler(RequestValidationError)
    async def validation_exception_handler(request, exc):
        return JSONResponse({"error":str(exc)}, status_code=400)
```

neomorphic (approved):
The frontend changes look good. There are probably many more cases (~25) where the response.json() should be caught to prevent this issue elsewhere. It looks like there are a number of instances later on in `frontend/src/queries/fileQueries.ts` that simply return an empty object if the json() method fails.

```
const body = await response.json().catch(() => ({}));
```

allison-truhlar:
> The frontend changes look good. There are probably many more cases (~25) where the response.json() should be caught to prevent this issue elsewhere. It looks like there are a number of instances later on in `frontend/src/queries/fileQueries.ts` that simply return an empty object if the json() method fails.
> 
> ```
> const body = await response.json().catch(() => ({}));
> ```
@neomorphic With the PR off this branch that addressed your above comment completed and merged, is this PR okay to merge into main now?

PR #289, opened 2026-01-07
fix: server error responses and display

allison-truhlar:
@neomorphic 
This PR adds the suggested general exception handler to the server to ensure any unhandled exceptions return a JSON response, and that the response includes the exception type and message. It also adds query utility functions to check for valid JSON in server responses, standardize the error that's thrown when the response is not okay, and encapsulate the pattern for many query mutations of sending a fetch request and then only checking the body/throwing an error if the response is not okay.

### DIFF
--- a/frontend/src/queries/preferencesQueries.ts
+++ b/frontend/src/queries/preferencesQueries.ts
@@ -112,12 +112,11 @@ const fetchPreferences = async (
   const data = await getResponseJsonOrError(response);
   if (response.ok) {
     return data as PreferencesApiResponse;
+  }
+  if (response.status === 404) {
+    return {}; // No preferences found, return empty object
   } else {
-    if (response.status === 404) {
-      return {}; // No preferences found, return empty object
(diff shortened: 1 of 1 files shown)


frontend/src/queries/preferencesQueries.ts
```diff
@@ -102,22 +106,18 @@ export const preferencesQueryKeys = {
...
-      { signal }
-    );
-    if (!response.ok && response.status === 404) {
+  const response = await sendFetchRequest('/api/preference', 'GET', undefined, {
+    signal
+  });
+  const data = await getResponseJsonOrError(response);
+  if (response.ok) {
+    return data as PreferencesApiResponse;
+  } else {
```
neomorphic: I don't think this else is necessary. You can just have the next if block, since the previous if check will return from the function if successful.

neomorphic (approved):
After much review offline, This looks good to me.

PR #290, opened 2026-01-07
SSH key manager

krokicki:
This PR adds an SSH key manager to the profile menu in the upper right. This lets you create SSH keys and authorize them so that you can seamlessly move between cluster nodes. The main purpose for this is to allow easy on boarding for our Seqera Platform instance.

I tried to be careful about making backups of files in ~/.ssh before changing them, but would welcome feedback especially if there are any edge cases I didn't think of. If you are testing this, I would recommend making a backup of your ~/.ssh directory first, just in case. 

@JaneliaSciComp/fileglancer @StephanPreibisch

### DIFF
--- a/frontend/src/components/ui/SSHKeys/GenerateTempKeyDialog.tsx
+++ b/frontend/src/components/ui/SSHKeys/GenerateTempKeyDialog.tsx
@@ -47,7 +47,7 @@ export default function GenerateTempKeyDialog({
         Generate SSH Key
       </Typography>
 
-      <Typography className="mb-4">
+      <Typography className="text-foreground mb-4">
         This will create a new ed25519 SSH key pair and add the public key to
         your authorized_keys file.
       </Typography>
@@ -63,7 +63,7 @@ export default function GenerateTempKeyDialog({
           type="password"
           value={passphrase}
         />
-        <Typography className="text-muted text-xs ml-1 mt-1">
+        <Typography className="text-foreground text-xs ml-1 mt-1">
           A passphrase adds extra security but must be entered each time you use
           the key.
         </Typography>
--- a/frontend/src/components/ui/SSHKeys/TempKeyDialog.tsx
+++ b/frontend/src/components/ui/SSHKeys/TempKeyDialog.tsx
@@ -59,7 +59,7 @@ export default function TempKeyDialog({
           <Typography className="text-warning font-semibold text-md">
             Copy this private key now - it will not be available again!
           </Typography>
-          <Typography className="mt-1">
+          <Typography className="text-warning mt-1">
             The private key is not stored anywhere. You must copy it now by
             clicking the "Copy Private Key" button and then save it somewhere
             safe (e.g. paste it into Seqera Platform's 'Managed Identities'


krokicki:
I added a `VITE_ENABLE_SSH_KEYS` feature toggle, and enabled both this and `VITE_ENABLE_TASKS` in the CI build. I also modified the implementation to do as little as possible with `~/.ssh`.

krokicki:
@mkitti As discussed, we agree about "treating .ssh as read-only as much as possible". To minimize complexity and because for now this feature has a very small user base, I've modified the GUI to do the most minimal thing possible to `~/.ssh`, namely: if `id_ed25519` does not exist, this tool can create one for you and add it to `authorized_keys` using `cat`. I've removed functionality around managing multiple keys, and all functionality that deletes keys. This minimizes risks to user's existing SSH keys and setups.

mkitti:
In a few places we still read the private key into resident memory of the Python process as an immutable Python string. This is not good. Core dumps of the Python process may contain the contents of the private key.

Rather than loading file contents into an immutable string, we should read the contents into a mutable `bytearray` and then explicitly overwrite the contents of the `bytearray` when the contents are no longer needed.

```python
import os

key_path = "private_key.pem"
file_size = os.path.getsize(key_path)

# 1. Pre-allocate mutable memory
sensitive_data = bytearray(file_size)

# 2. Read directly into the pre-allocated buffer
with open(key_path, "rb") as f:
    f.readinto(sensitive_data)

# Use the bytes as needed

# 3. Securely overwrite the memory with zeros
for i in range(len(sensitive_data)):
    sensitive_data[i] = 0

# 4. Now safe to remove the reference
del sensitive_data
```

References:
1. https://stackoverflow.com/questions/728164/securely-erasing-password-in-memory-python#:~:text=The%20correct%20solution%20is%20to,bytearray%22%20instead%20of%20python%20strings.
2. https://softwareengineering.stackexchange.com/questions/270113/using-a-bytearray-rather-than-a-string-to-store-password-in-memory#:~:text=Using%20a%20bytearray%20datatype%20to,in%20python%203.4%20using%20bytearray?

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,444 @@
...
+
+class GenerateKeyResponse(BaseModel):
+    """Response after generating an SSH key"""
+    key: SSHKeyInfo = Field(description="The generated key info")
+    message: str = Field(description="Status message")
+
+
+class SSHKeyContent(BaseModel):
+    """SSH key content - only fetched on demand"""
+    key: str = Field(description="The requested key content")
```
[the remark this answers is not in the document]
mkitti: Perhaps we could make `SSHKeyContent` follow the context manager protocol so we can use a `with` statement. On `__exit__` the context manager would then make sure to shred the key contents in memory by zeroing out the resident `bytearray`. You could use [`contextlib`](https://docs.python.org/3/library/contextlib.html#contextlib.AbstractContextManager) to help with this.

I also suggest centralizing the the potentially sensitive reading implementation here.

1. The class would only store a string or Path containing the location of the key file.
2. The context manager `__enter__` would read the file contents into a `bytearray` via `readinto` and then return that `bytearray`
3. The context manager `__exit__` would then explicitly overwrite the `bytearray`.

This will minimize the time the any sensitive secret spends in resident memory.

fileglancer/app.py
```diff
@@ -862,6 +863,115 @@ async def get_profile(username: str = Depends(get_current_user)):
...
+        username: str = Depends(get_current_user)
+    ):
+        """Get the public or private key content for copying"""
+        if key_type not in ("public", "private"):
+            raise HTTPException(status_code=400, detail="key_type must be 'public' or 'private'")
+
+        with _get_user_context(username):
+            try:
+                ssh_dir = sshkeys.get_ssh_directory()
+                return sshkeys.get_key_content(ssh_dir, "id_ed25519", key_type)
```
mkitti: Shred the key contents in a bytearray after usage. Ensure the contents get shreded even if an error occurs. Consider using a context manager here.

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,444 @@
...
+                if f.tell() > 0:  # File is not empty
+                    f.seek(-1, 2)  # Seek to last byte
+                    if f.read(1) != b'\n':
+                        # File doesn't end with newline, add one
+                        with open(authorized_keys_path, 'a') as af:
+                            af.write('\n')
+
+        # Append the key
+        with open(authorized_keys_path, 'a') as f:
+            f.write(public_key + '\n')
```
mkitti: Could we add a `restrict` keyword here? Does Seqera need an interactive terminal or port forwarding capabilities?

https://manpages.debian.org/experimental/openssh-server/authorized_keys.5.en.html#restrict

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,486 @@
...
+        with open(pubkey_path, 'r') as f:
+            return SSHKeyContent(key=f.read().strip())
+
+    elif key_type == "private":
+        private_key_path = safe_join_path(ssh_dir, filename)
+        if not os.path.exists(private_key_path):
+            raise ValueError(f"Private key '{filename}' not found")
+        # Use secure reading for private keys
+        key_content = _read_file_secure(private_key_path)
+        return SSHKeyContent(key=key_content)
```
mkitti:
```suggestion
        return SSHKeyContent(
            key=_read_file_secure(private_key_path)
        )
```

Remove extra references to the string.

mkitti (changes requested):

krokicki:
@mkitti @clements @allison-truhlar 
This PR is ready to be reviewed, thanks!

frontend/src/components/ui/Dialogs/NGLinkDialog.tsx
```diff
@@ -248,13 +248,13 @@ export default function NGLinkDialog({
 
   return (
     <FgDialog onClose={resetAndClose} open={open}>
-      <div className="mt-8 flex flex-col gap-2">
-        <Typography className="text-foreground font-semibold" type="h6">
-          {isEditMode
-            ? 'Edit Neuroglancer Short Link'
-            : 'Create Neuroglancer Short Link'}
-        </Typography>
+      <Typography className="text-foreground font-semibold text-lg mb-4">
```
mkitti: Is this an unrelated UI change?

frontend/src/components/ui/Table/TableCard.tsx
```diff
@@ -62,7 +62,7 @@ declare module '@tanstack/react-table' {
       data: CellContextMenuData
     ) => void;
   }
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mkitti: I think this came in with one of my commits, but it may be unrelated.

frontend/src/components/ui/SSHKeys/TempKeyDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+          SSH Key Generated
+        </Typography>
+      </div>
+
+      <div className="space-y-4">
+        <div className="p-3 bg-warning/10 border border-warning/20 rounded-md">
+          <Typography className="text-warning font-semibold text-md">
+            Copy this private key now - it will not be available again!
+          </Typography>
+          <Typography className="mt-1">
```
neomorphic: Adding a text-warning class here makes it look nicer in dark mode

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,560 @@
...
+        ssh_dir: Path to the .ssh directory
+    """
+    if not os.path.exists(ssh_dir):
+        os.makedirs(ssh_dir, mode=0o700)
+        logger.info(f"Created SSH directory: {ssh_dir}")
+    else:
+        current_mode = os.stat(ssh_dir).st_mode & 0o777
+        if current_mode != 0o700:
+            os.chmod(ssh_dir, 0o700)
+            logger.info(f"Fixed SSH directory permissions: {ssh_dir}")
```
neomorphic: We are changing the permissions of the users .ssh directory, but not telling them. Should we inform them as part of the process as a public service?

frontend/src/components/ui/SSHKeys/TempKeyDialog.tsx
krokicki: Fixed.

frontend/src/components/ui/SSHKeys/GenerateTempKeyDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  };
+
+  return (
+    <FgDialog onClose={handleClose} open={showDialog}>
+      <Typography className="text-foreground font-semibold text-lg mb-4">
+        Generate SSH Key
+      </Typography>
+
+      <Typography className="mb-4">
```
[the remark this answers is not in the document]
krokicki: Fixed.

fileglancer/sshkeys.py
krokicki: Not a bad idea, but I'm not sure it's worth worrying about.

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,560 @@
...
+    try:
+        type_idx = next(i for i, part in enumerate(parts) if part.startswith(SSH_KEY_PREFIX))
+    except StopIteration:
+        raise ValueError("Invalid public key format: key type not found")
+
+    # Handle options
+    if type_idx > 0:
+        options = parts[0].split(',')
+        if "restrict" not in options:
+            options.insert(0, "restrict,pty")
```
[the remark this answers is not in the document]
krokicki: I don't have time to dig into this now, if you have improvements along these lines, please submit them as a new PR once this is merged. On the other hand, I'm really not sure this is worth optimizing.

fileglancer/sshkeys.py
```diff
@@ -0,0 +1,560 @@
...
+    temp_key_path = os.path.join(temp_dir, "temp_key")
+    temp_pubkey_path = os.path.join(temp_dir, "temp_key.pub")
+
+    old_umask = os.umask(0o077)
+    try:
+        passphrase_str = passphrase.get_secret_value() if passphrase else ""
+        cmd = [
+            'ssh-keygen',
+            '-t', 'ed25519',
+            '-N', passphrase_str,
```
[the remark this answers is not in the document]
krokicki: I noticed that too. Doesn't seem there is much that can be done.

frontend/src/components/ui/Table/TableCard.tsx
krokicki: Unrelated to what? In any case, it doesn't seem necessary.

PR #297, opened 2026-01-15
Added Devcontainer

krokicki:
I'm adding a devcontainer we can use to run a containerized version of Claude (and potentially other AI agents in the future). Please try it out and let me know if you run into any issues.

One thing that isn't working quite right are the UI tests. I installed all the dependencies and it mostly works, but it's still failing one test and it hangs at the end waiting for the browser to close. 

I also added a CLAUDE.md that we can iterate on.

### DIFF
--- a/docs/DevContainer.md
+++ b/docs/DevContainer.md
@@ -141,7 +141,7 @@ pixi run container-claude
 
 # Or from inside a container shell
 pixi run container-shell
-claude --dangerously-skip-permissions
+pixi run claude --dangerously-skip-permissions
 ```
 
 ### Stop the Container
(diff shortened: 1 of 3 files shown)


neomorphic:
@krokicki I tried to run this on my dev machine and encountered an error when building the dev container. I have somewhat restrictive permissions on my .claude directory and the container fails to build, because it doesn't have access to that directory. Did you encounter this on your linux box or are you using a setup that has your home directory mounted on a local drive instead of in /groups/*****?

krokicki:
@neomorphic Following up on Slack.

PR #301, opened 2026-01-21
test: combined action no pixi cache

allison-truhlar:


allison-truhlar:
Closing this PR because the existing workflow of running the build and integration testing jobs concurrently is faster than running them in sequence.

PR #302, opened 2026-01-22
feat: SSHKeyContentResponse, SecretStr, authorized_keys options, and tests

mkitti:
* Add custom SSHKeyContentResponse that wipes memory after sending content
* SSHKeyContentResponse also sends the response line by line to minimize caching downstream
* Use `pydantic.SecretStr` to guard ssh private key password
* Add `restrict` and `pty` keyword to authorized_keys options. Consider whether Seqera needs a `pty` or not.
* Add `fileglancer` to authorized_keys comment
* Only manage keys which have fileglancer in the comment. New keys have this comment.
* Add a "temporary" key option which the permanent key is not saved to the server and shown once to the user. Only the public key is added to authorized_keys. This is useful if the user already has an unmanaged "permanent" key but still needs to generate a key via this interface.
* The only "permanent" key is the default id_ed25519 option. This only managed if it has a "fileglancer" comment.
* Added the ability to regenerate id_ed25519.pub if only id_ed25519 exists.
* Add tests for SSH key handling

image

Questions:
1. The naming of "permanent" versus "temporary" keys may be confusing. It might be miscontrued to mean that the "temporary" key only works for a limited time whereas it is merely just not saved. Is there better terminology?
2. Do we actually need to generate the private key on the server side?
3. How can we best balance convenience with the current [zero-trust security](https://www.appgate.com/products/zero-trust-network-access) framework of HHMI central IT?

### DIFF
--- a/tests/test_sshkeys.py
+++ b/tests/test_sshkeys.py
@@ -1,6 +1,7 @@
 "Tests for SSH key management utilities with secure bytearray handling."
 
 import os
+import stat
 import subprocess
 import tempfile
 import pytest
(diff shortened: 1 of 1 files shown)


mkitti:
Note that this pull request is into the `ssh-key-manager` branch, rather than main. The sequence would be to first merge this, and then return to #290 .

PR #306, opened 2026-01-28
feat: Initial Zip file and OME-Zarr Archive (RFC-9) support

mkitti:
# Purpose.

[Support OME-Zarr zip archives (RFC-9)](https://ngff.openmicroscopy.org/rfc/9/index.html). I am an author of OME-Zarr RFC-9. Zarr zip file implementations are wide spread and implemented in a number of Zarr implementations.

Notably Neuroglancer implements [Zip file key-value](https://neuroglancer-docs.web.app/datasource/zip/index.html) store.

# Background

Zip files are documented by a [PKWare appnote](https://support.pkware.com/pkzip/appnote). A zip file contains local file entries followed by a central directory at the end of the file. While the local file entries at the beginning file could be read in streaming fashion, it is often necessary to read the central directory at the end of the file first. Duplicate local file entries may exist in the beginning of the archive, but only the central directory can correctly indicate the latest version for example.

OME Zarr RFC-9 proposes a standard way to store OME-Zarr datasets in zip files with a OZX extension. A particular challenge for Zarr in zip files in the large numbers of files which may make the central directory more difficult to parse. The RFC recommends a number of recommendations to minimize the number of files such as using sharding.

Another recommendation by the RFC is to list zarr.json metadata first in the central directory. This allows an application like fileglancer which is mainly concerned with parsing metadata to quickly access the metadata without having the parse the entire central directory. It also allows the entire tree of the hieararchy to be elucidated. To facilitate detection of this optimization, JSON is stored in the comment of the zip which occurs at the end of the central directory. This contains a `jsonFirst` flag to indicate that the writer of the file placed the JSON files first after the central directory. If this flag is true, a reader may assume that no further zarr.json files exist within the archive once the first non-zarr.json file is read.

# Design

While Python does provide a [zipfile](https://docs.python.org/3/library/zipfile.html) standard library, a key performance optimization here is the abilty to read a partial central directory. While browsing, only the metadata may be needed. As such this pull request contains new code to read the central directory partially, which the standard library does not support.

If there becomes a need stream the contents of the zip file for external applications, then some additional reading and caching of the central directory may be needed. 

While I was initially working on this feature with OZX in mind, implementing generic zip file support also seemed useful. Thus the core implementation splits generic zip support and then builds OME-Zarr support on top of that.

### DIFF
--- a/docs/RFC-9-OZX-Implementation.md
+++ b/docs/RFC-9-OZX-Implementation.md
@@ -26,9 +26,9 @@ This document describes the implementation of RFC-9 support for reading OME-Zarr
 ┌─────────────────────────────────────────────────────────────────┐
 │                         Backend                                 │
 │  ┌──────────────────────┐    ┌─────────────────────────────┐    │
-│  │ /api/ozx-content/    │    │ OZXReader (ozxzip.py)       │    │
+│  │ /api/zip-content/    │    │ OZXReader (ozxzip.py)       │    │
 │  │ /api/ozx-metadata/   │───▶│ - OME metadata parsing      │    │
-│  │ /api/ozx-list/       │    │ - jsonFirst optimization    │    │
+│  │ /api/zip-list/       │    │ - jsonFirst optimization    │    │
 │  └──────────────────────┘    ├─────────────────────────────┤    │
 │                              │ ZipReader (zipread.py)      │    │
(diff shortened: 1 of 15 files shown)


mkitti:
Closing in favor of https://github.com/JaneliaSciComp/zipglancer

PR #308, opened 2026-02-04
fix: do not load large binary files on frontend

allison-truhlar:
Clickup id: 86ae03g9k

Prevents large binary files from being fully downloaded when viewing in the file preview panel. The backend now detects whether a file is binary by analyzing the first 4KB using the same control character heuristic that was previously implemented on the frontend on already-fetched file content. The frontend uses this information to skip content fetching for binary files. 

To verify:
- Open the developer tools network tab and navigate to /nrs/opendata/.DS_Store in Fileglancer. Filter for `/api/content/` - you'll only see a HEAD request for the file, no GET request, and that the response headers include `x-is-binary: true`.
- Compare to /nrs/opendata/data_external/README. This file has both a HEAD request and a GET request, since the HEAD request returns `x-is-binary: false`.

**Changes**

- `fileglancer/utils.py`: Added `is_likely_binary()` function that analyzes bytes for control characters (>1% threshold indicates binary) - this is the same heuristic that was previously implemented on the frontend
- `fileglancer/app.py`: Updated HEAD `/api/content` endpoint to read first 4KB, run binary detection, and return `X-Is-Binary` header
- `frontend/src/queries/fileContentQueries.ts`: Added `useFileMetadataQuery` hook to fetch file metadata via HEAD request; simplified `useFileContentQuery` by removing client-side binary detection (now done on the backend)
- `frontend/src/components/ui/BrowsePage/FileViewer.tsx`: Updated to call HEAD endpoint first and only fetch content if file is not binary; displays "Binary file - preview not available" message for binary files
- `frontend/src/utils/index.ts`: Added HEAD method support to `sendFetchRequest`

**Tests**

- Added unit tests for `is_likely_binary()` covering text files and binary files
- Added tests for HEAD endpoint verifying correct `X-Is-Binary` header for text, JSON, binary, large files, directories, and empty files

@krokicki

### DIFF
--- a/fileglancer/filestore.py
+++ b/fileglancer/filestore.py
@@ -38,7 +38,6 @@ class FileInfo(BaseModel):
     absolute_path: Optional[str] = None
     size: int
     is_dir: bool
-    is_binary: Optional[bool] = None
     permissions: str
     owner: Optional[str] = None
     group: Optional[str] = None


neomorphic (approved):
This worked when I tested it and the code looks good to me.

krokicki:
Nice work, @allison-truhlar! Really glad to see this fix, I keep tripping over this bug and crashing my browser tab.

fileglancer/app.py
```diff
@@ -1124,8 +1124,23 @@ async def head_file_content(path_name: str,
...
 
+                # Only check binary status for files, not directories
+                is_binary = False
+                if not file_info.is_dir:
+                    try:
+                        # Read first 4KB to determine if file is binary, using validated path
+                        full_path = os.path.realpath(file_info.absolute_path)
+                        with open(full_path, 'rb') as f:
+                            sample = f.read(4096)
+                            is_binary = is_likely_binary(sample)
```
krokicki: We may want to move this binary checking stuff into the Filestore class, so that it can be abstracted later. Other implementations (e.g. S3) might have different ways to implement this.

PR #309, opened 2026-02-04
feat: recognize symbolic links

allison-truhlar:
Clickup id: 86aewngc6

This PR adds full support for symlinks in the file browser of Fileglancer, with visual distinction and navigation. Users can identify symlinks by a link icon and "Symlink" type label in the file table and the properties panel. Broken symlinks show a broken link icon. Clicking on a working symlink navigates to its target location within the same or different file share.

This PR also contains two related, but not symlink functionality/display only, changes:
1. Data link creation now correctly targets the selected item in the file browser, if applicable. Previously, creating a data link for a subdirectory or symlink selected from its parent would create the link for the parent directory instead. Viewer icons (Neuroglancer, etc.) always create data links for the current browsed directory, even when a subdirectory row is selected in the properties panel.
2. Pressing 'escape' now clears the selected row and reverts the properties panel data to the current directory.

To verify working symlink detection, display, and navigation:
- In the dev Fileglancer launched from this branch, navigate to `/nrs/cellmap/data/jrc_22ak351-leaf-2l`. You should see the N5 directory identified as a symlink. Clicking on the name, you can navigate to the symlinked path.
- If you navigate to the same directory in the production Fileglancer ([here](https://fileglancer.int.janelia.org/browse/nrs_cellmap/data/jrc_22ak351-leaf-2l)), the N5 directory is identified as a regular Folder.  If you click on the name, the file browser can't navigate to it. The breadcrumbs display the error "Error loading path" and eventually the file table shows the error "Folder not found."  

**Tests**
- Added tests in tests/test_filestore.py covering symlink detection, same-share/cross-share resolution, relative symlinks, broken symlink handling, and symlinks to directories
- Added Playwright tests for symlinks in frontend/ui-tests/tests/symlink-navigation.spec.ts covering icon display, navigation, type labeling, context menu, broken symlinks, and properties panel content 
- Added Playwright tests in frontend/ui-tests/tests/data-link-operations.spec.ts verifying data link targets the selected subdirectory (not parent) and viewer icon targets the current directory (not selected row)

@krokicki

### DIFF
--- a/frontend/src/components/ui/BrowsePage/FileBrowser.tsx
+++ b/frontend/src/components/ui/BrowsePage/FileBrowser.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import type { MouseEvent } from 'react';
 import { Typography } from '@material-tailwind/react';
 import toast from 'react-hot-toast';
@@ -52,7 +53,8 @@ export default function FileBrowser({
     fspName,
     fileQuery,
     fileBrowserState,
-    updateFilesWithContextMenuClick
+    updateFilesWithContextMenuClick,
+    clearSelection
   } = useFileBrowserContext();
   const { folderPreferenceMap, handleContextMenuFavorite } =
     usePreferencesContext();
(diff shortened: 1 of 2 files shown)


allison-truhlar:
Note: commit https://github.com/JaneliaSciComp/fileglancer/pull/309/commits/08fff7faa182b66906c334b95c82b2e376f6f6b5 adds tests for these cases we discussed - can you access sensitive data from a data link to a directory that contains a symlink to a:
-  sensitive directory/file?
- another symlink to a sensitive directory/file (ie., symlink -> symlink -> target)?

It simulates a data link by using /api/content/.

PR #314, opened 2026-02-15
Fileglancer Apps

krokicki:
This PR adds **Apps** which are CLI tools that can be executed on the cluster from a web form, and monitored within Fileglancer. 

There are two new top level pages: **Apps** and **Jobs**. The Apps page allows you to add apps via a GitHub URL that conforms to a new "Runnables Protocol" that I created based on inspiration from [Nexflow Schema](https://nf-co.re/docs/nf-core-tools/pipelines/schema), [CWL](https://www.commonwl.org/), [IPP](https://github.com/JaneliaSciComp/image_processing_pipeline/), and [Fractal](https://fractal-analytics-platform.github.io/tasks_spec/#task-list-and-manifest). In short, each app repo contains a `runnables.yaml` manifest that declares one or more entry points, and defines their parameters, resource defaults, and environment variables. A single repo can contain multiple manifests in subdirectories, each discovered and registered as a separate app. Manifests can also reference a separate `repo_url` for the actual tool code. This will allow us to create a "catalog" repo of external tools in cases their tool's repo can't host the `runnables.yaml`. 

Currently, the prerequisites support Pixi, Maven, and NPM, and only Pixi has been tested. I'd like to expand this to also include Conda and containers. Allowing uncontainerized tools like Pixi is a bit of a risk from a complexity standpoint, and also won't work for users who don't have these prereqs set up. I'd like to keep thinking about the best solution to this, such as offering solutions for automatically installing prereqs, pre-installing all prerequisite platforms in a shared location during Fileglancer server setup, exploring solutions like [album](https://album.solutions/), etc.

App URLs are stored in your preferences and the repos are cloned and cached in `~/.fileglancer/apps`. When you launch an entry point, the manifest is used to auto-generate a web form with appropriate controls for each parameter type. Parameters can be organized into collapsible sections. File and directory parameters use the file selection dialog.

Jobs are submitted to the cluster via a new [py-cluster-api](https://github.com/JaneliaSciComp/py-cluster-api) package (inspired by [dask-jobqueue](https://github.com/dask/dask-jobqueue) and our [java-lsf](https://github.com/JaneliaSciComp/java-lsf)), which initially supports local and LSF executors. We can easily add others (e.g. Slurm) as necessary for non-Janelia users. The cluster executor type and default resources are configured in Fileglancer's `config.yaml` under a new `cluster` section. See `docs/config.yaml.template` for the possible options. 

The Jobs page lists all submitted jobs and updates periodically. Clicking on a Job brings you to a Job Detail page that shows four tabs: parameters, the generated script, stdout, and stderr. Jobs can be cancelled, deleted, or relaunched with the same (or modified) parameters.

On server restart, the job monitor automatically reconnects to any active cluster jobs and resumes polling, as long as the `job_name_prefix` is set (otherwise this is determined randomly each time, similar to the `session_secret_key`). A reconciliation loop detects zombie jobs and syncs status transitions from the executor to the database.

## Services 

Services are cluster jobs that are started and stopped by the user as needed. They can be web servers, notebooks, APIs, etc. 
* Entry points can declare `type: service` in `runnables.yaml`. These run until explicitly stopped rather than running to completion.
* Fileglancer exports `SERVICE_URL_PATH` as an env var in the job script. Services write their URL to a new file at `SERVICE_URL_PATH`. Fileglancer polls this file and when the URL is written it gets displayed in the UI for the user to access.

## Trying it out

You can paste in these app URLs:
* Demo app: https://github.com/JaneliaSciComp/fileglancer-demo-app
* MRC viewer app: https://github.com/JaneliaSciComp/neuroglancer-mrc
* bioformats2raw: https://github.com/JaneliaSciComp/nf-omezarr (branch `fileglancer`) 
* Cellmap Flow: https://github.com/janelia-cellmap/cellmap-flow (branch `fileglancer`)

@StephanPreibisch @JaneliaSciComp/fileglancer @dchen116 @mzouink

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,9 +25,9 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Install pixi
-      uses: prefix-dev/setup-pixi@v0.9.1
+      uses: prefix-dev/setup-pixi@v0.9.4
       with:
-        pixi-version: v0.55.0
+        pixi-version: v0.65.0
         cache: false
         manifest-path: pyproject.toml
 
@@ -81,9 +81,9 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Install pixi
-      uses: prefix-dev/setup-pixi@v0.9.1
+      uses: prefix-dev/setup-pixi@v0.9.4
       with:
-        pixi-version: v0.55.0
+        pixi-version: v0.65.0
         cache: false
         manifest-path: pyproject.toml
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -212,9 +212,9 @@ cd frontend/ui-tests && pixi run npx playwright install
 ## Technology Stack
 
 ### Backend
-- Python 3.14+
+- Python 3.12+
 - FastAPI
-- SQLAlchemy (async with aiosqlite/psycopg2)
+- SQLAlchemy (synchronous)
 - Alembic (migrations)
 - Pydantic (validation)
 - Uvicorn (ASGI server)
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -9,10 +9,16 @@ git clone git@github.com:JaneliaSciComp/fileglancer.git
 cd fileglancer
 ```
 
-If this is your first time installing the extension in dev mode, install package in development mode.
+Install the package in development mode.
 
 ```bash
 pixi run dev-install
+```
+
+Copy the configuration file and edit as desired (more details in [Configuration](#configuration) below).
+
+```bash
+cp docs/config.yaml.template config.yaml
 ```
 
 Now you can launch the server. It will automatically rebuild when there are file changes to the backend.
@@ -31,11 +37,7 @@ View the app in the browser at localhost:7878.
 
 ## Configuration
 
-Copy the configuration file and edit as desired.
-
-```
-cp docs/config.yaml.template config.yaml
-```
+The `config.yaml` file at the project root provides the server configuration.
 
 ### File Share Paths
 
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -1,5 +1,13 @@
 # Making a new release of Fileglancer
 
+## Run Python 3.12 compatibility tests
+
+Since the PyPI package supports Python 3.12 (`requires-python = ">=3.12.0"`), verify that the backend tests pass under Python 3.12 before releasing:
+
+```bash
+pixi run test-py312
+```
+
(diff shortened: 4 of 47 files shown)


docs/AuthoringApps.md
```diff
@@ -0,0 +1,617 @@
...
+        type: file
+        required: true
+```
+
+When `conda_env` is set, the generated script initializes conda and activates the environment before any env vars, pre_run, or the main command:
+
+```bash
+eval "$(conda shell.bash hook)"
+conda activate my-analysis-env
+```
```
mkitti: I'm confused why a service would go this route.

1. `conda run` is generally the recommended way to run a specific command within an environment, especially during non-interactive use. `conda activate` is mainly intended for interactive use.

https://docs.conda.io/projects/conda/en/stable/commands/run.html

2. If the main purpose is to just to run Python, often this can be done by just providing the full path to a Python executable since conda-forge hardcodes the library path in the executable. This is how Jupyter kernels within conda environments generally function. I would consider an example that more strongly depends on environment activation.

docs/AuthoringApps.md
```diff
@@ -0,0 +1,617 @@
+# Authoring Apps for Fileglancer
+
+Fileglancer can discover and run apps from GitHub repositories. An app is defined by a `runnables.yaml` manifest file that describes one or more commands (called **runnables**) that users can launch as cluster jobs through the Fileglancer UI.
+
+## Quick Start
+
+1. Create a `runnables.yaml` file in your GitHub repository
```
mkitti: While I can imagine a few reasons why you are using YAML here, I am surprised to see YAML used for a new project. The pitfalls of YAML are well documented now.

https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

Would plain JSON or small extension of it (i.e. JSONC, JSON5) ve sufficient?

krokicki: YAML has its issues, but so does every other format. I don't think any of the problems in that blog are relevant to how we're using YAML here. 

I experimented with both JSON and TOML and I didn't like how the files looked for this use case. JSON is too verbose and difficult to edit (lived experience with `nextflow_schema.json`) and TOML is not great for lists of objects.

docs/AuthoringApps.md
krokicki: Sure, this was just a quick hack to get things working with Cellmap Flow. We can revisit this later, if we want to continue supporting conda. Right now, I'm not even sure if that's a good idea.

PR #335, opened 2026-03-20
Further fixes to job submission

allison-truhlar:
This PR troubleshoots submitting jobs to the Janelia cluster when running the Fileglancer server as root.

**Changes so far:**
1. [Work directory fix](https://github.com/JaneliaSciComp/fileglancer/commit/543d9cb7fd17b28a61e93bb138b1cada2ec8fd1e) - Resolved ~ to the actual user's home (~username), not root's home.
2. [seteuid job submission](https://github.com/JaneliaSciComp/fileglancer/commit/3a016beaeefa7aac0d648872d26e56850611b0ea) - Wrapped work dir creation, symlink, and executor.submit() in the EffectiveUserContext so filesystem operations happen as the user.
3. [Stale symlink fix](https://github.com/JaneliaSciComp/fileglancer/commit/95742f461e1b9caf36bb24405c2729de96f629a5) - Unlink existing repo symlink before creating a new one (caused an issue when a job failed - the symlink wasn't removed).
4. [bsub -U username flag](https://github.com/JaneliaSciComp/fileglancer/commit/99fda28323be5d6b59e25e9945e3bd996168f740) — Because bsub uses getuid() (real UID = root) not geteuid(), we pass -U username to tell LSF to run the job as the actual user. Note: unsure whether this is actually needed after adding the missing FGC_CLUSTER__EXECUTOR=lsf to the fileglancer-hub .env file — should test removing it once everything else is fixed
5. [cluster.extra_paths config option](https://github.com/JaneliaSciComp/fileglancer/commit/95b301ff269e68823639bda63cf1116d8378f95b) — Prepends directories to os.environ["PATH"] at server startup so bsub/bjobs/bkill are findable. Needed because systemd services have minimal PATH. Documented in config.yaml.template.

**Where we are now** (as of release 2.7.0a7, commit eff3c098a1ed08614fedaa4a74ba68cd0333c273)
- bsub is found on PATH (no more FileNotFoundError), but it fails because LSF can't find its configuration

@krokicki

### DIFF
--- a/docs/config.yaml.template
+++ b/docs/config.yaml.template
@@ -142,6 +142,8 @@ cluster:
 #     - /opt/lsf/bin           # startup so scheduler commands (bsub, bjobs,
 #                              # bkill) are findable. Useful when running as a
 #                              # systemd service with a minimal default PATH.
+#   extra_env:                 # Extra environment variables set at startup.
+#     LSF_ENVDIR: /misc/lsf/conf  # Required for LSF to find lsf.conf.
 #   job_name_prefix: fg        # Prefix for cluster job names. REQUIRED for job
 #                              # reconnection after server restarts. Without this,
 #                              # active jobs will not be re-tracked and will
--- a/docs/config.yaml.template
+++ b/docs/config.yaml.template
@@ -138,6 +138,10 @@ session_cookie_secure: true
 #
 cluster:
   executor: local            # "local" or "lsf"
+#   extra_paths:               # Directories prepended to the server's PATH at
+#     - /opt/lsf/bin           # startup so scheduler commands (bsub, bjobs,
+#                              # bkill) are findable. Useful when running as a
+#                              # systemd service with a minimal default PATH.
 #   job_name_prefix: fg        # Prefix for cluster job names. REQUIRED for job
 #                              # reconnection after server restarts. Without this,
 #                              # active jobs will not be re-tracked and will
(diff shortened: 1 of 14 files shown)


krokicki:
The rationale for [this commit](https://github.com/JaneliaSciComp/fileglancer/pull/335/changes/d5070516335cb8c204bdc5ba3643a1f74b5cf573) is concerning. It moves file access to the worker.py because:

> EffectiveUserContext sets process-wide euid/egid, which gets corrupted when
concurrent async requests interleave at await points. This caused permission
errors for users with restrictive home directory permissions (drwx------) and
setegid failures ("Operation not permitted") for other concurrent users.
> Replace in-process EffectiveUserContext usage for all apps/manifest operations
with worker subprocesses via _run_as_user, which sets the real UID/GID and
isolates each user's identity in a separate process. Also set HOME in the
worker environment so git finds the correct user config.

Does this mean that the EffectiveUserContext doesn't work for certain cases? What are those cases exactly? I can view my `.ssh` directory which has 700 permissions.

allison-truhlar:
> The rationale for [this commit](https://github.com/JaneliaSciComp/fileglancer/pull/335/changes/d5070516335cb8c204bdc5ba3643a1f74b5cf573) is concerning. It moves file access to the worker.py because:
> 
> > EffectiveUserContext sets process-wide euid/egid, which gets corrupted when
> > concurrent async requests interleave at await points. This caused permission
> > errors for users with restrictive home directory permissions (drwx------) and
> > setegid failures ("Operation not permitted") for other concurrent users.
> > Replace in-process EffectiveUserContext usage for all apps/manifest operations
> > with worker subprocesses via _run_as_user, which sets the real UID/GID and
> > isolates each user's identity in a separate process. Also set HOME in the
> > worker environment so git finds the correct user config.
> 
> Does this mean that the EffectiveUserContext doesn't work for certain cases? What are those cases exactly? I can view my `.ssh` directory which has 700 permissions.

Followed up on this via Slack.

PR #338, opened 2026-03-24
feat: add search input and apply group display preferences to file selector

allison-truhlar:
Clickup id: 86afydrnp

This PR improves the FileSelector component by adding search/filtering capabilities at all navigation levels, applying the group display preference for filtering the initial zones displayed, and aligning its visual styling with the file browser for a consistent user experience. 

This PR also fixes:
- 'local' group being excluded from file share provider visibility filtering
- the file selector not opening to the correct file system location if a path format other than Linux was used

@krokicki


fileglancer/apps/core.py
```diff
@@ -515,6 +522,21 @@ def _validate_parameter_value(param: AppParameter, value, session=None) -> str:
 
     if param.type in ("file", "directory"):
         str_val = str_val.replace("\\", "/")
+        # Resolve Mac (smb://), Windows (//server/share), or alternate Linux
+        # paths to the server's mount_path so validation and shell commands
+        # use the canonical server path.
+        if session is not None:
+            from fileglancer.database import resolve_any_path_format
+            resolved = resolve_any_path_format(session, str_val)
```
krokicki: Does the backend really need to know how to translate these other formats? Why not just translate on the client side and always send a Linux path to the backend? That would reduce the amount of code we need to maintain.

krokicki (commented):
Looks good, I just have a question about frontend/backend.

It would be nice if the dialog links worked the same way as the normal browser (navigate via single click on the name), but maybe that's a separate issue.

fileglancer/apps/core.py
allison-truhlar: Fixed with commit https://github.com/JaneliaSciComp/fileglancer/pull/338/commits/b046c9c7151aa142e6172ca4ba189bf735278ecc

PR #344, opened 2026-04-07
feat: add preference to show full file path in data links; add labels for UI display to data links

allison-truhlar:
Clickup ids: 86abc1mrr

This PR adds the ability for the user to select between three data link naming formats:
- **Directory name only** - current format, `https://.../<key>/directory-name`
- **Full/transparent path** — when enabled the full relative path appears in the URL instead of just the directory name. `https://.../<key>/full/path/to/directory-name`
- **Custom** - allows the user to type a custom string into an input field. `https://.../<key>/some-custom-string/can/be/path-like`

Whatever the string added for the data link name, this is added to the "name" column in the data link table. This value is not editable.


krokicki (commented):
I like the data labels!

For the transparent URLs, it shows me that the link will look like this:
`https://.../<key>/annotations/jrc_P3-E5-D1-N2_v2_destreaked/crop12/crop12cell_labels_for_ML/amira_export.zarr`

It needs to include the full path (i.e. mount path of the FSP at the beginning) to be fully transparent.

krokicki (approved):
Looks good. I think it would be better if the "Enable automatic data link creation" option was on the bottom, but that's not easy to do with how the options are currently implemented, so we can come back to it later.

PR #357, opened 2026-04-21
fix: map loguru log levels to stdlib equivalents in worker subprocess

allison-truhlar:
The parent process passes FGC_LOG_LEVEL (a loguru level) to the worker subprocess, which uses stdlib logging. Loguru-specific levels like TRACE (used in my local dev mode, which is where I discovered the bug) crash the worker with "Unknown level" ValueError, preventing app manifest fetches from completing.
@krokicki


krokicki:
Should we maybe switch the worker process to use loguru in the future?

But I know @neomorphic was worried about import size.

PR #363, opened 2026-04-29
feat: allow changing execute permission

allison-truhlar:
Clickup id: 86ah03cn7

This PR allows the user to change the execute permissions on files and folders from the properties drawer. It also adds the ability to change the "sticky bit" permission - whether only the owner can rename or delete files in a directory. The latter is displayed in a new table under the main permissions table in the properties drawer and can be changed via a new checkbox in the ChangePermissions dialog, for directories only.

@krokicki

### DIFF
--- a/fileglancer/filestore.py
+++ b/fileglancer/filestore.py
@@ -748,17 +748,23 @@ class Filestore:
             raise ValueError("Permissions must be a string of length 10")
         full_path = self._check_path_in_root(path)
         # Convert permission string (like '-rw-r--r--') to octal mode
+        # Execute positions (3, 6, 9) can contain special characters:
+        #   'x' = execute, 's'/'S' = setuid/setgid, 't'/'T' = sticky bit
+        #   Lowercase = execute set, uppercase = execute not set
         mode = 0
         # Owner permissions (positions 1-3)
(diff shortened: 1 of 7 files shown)


allison-truhlar:
@neomorphic @cgoina Here is what the properties panel looks like with group sticky bit line in the table.

What do you think of the wording? I feel torn about whether it would be more clear to use the technical language, if we assume the main users of these special permissions would be more technical, versus the less-technical language I'm using now to try to make it understandable to users not familiar with sticky bits (like myself prior to this PR):

Example where the directory has the same group ownership (scicompsoft) as the user's primary group (scicompsoft), and the group sticky bit isn't set:
Screenshot 2026-04-30 at 12 54 39 PM

Example where the directory has a different group ownership (opendata) than the user's primary group (scicompsoft) as well as the parent directory (scicompsoft), and the group sticky bit is set:
Screenshot 2026-04-30 at 12 54 57 PM

For completeness, here is what the option looks like in the change permissions dialog:
Screenshot 2026-04-30 at 12 55 15 PM

neomorphic:
I think a slight tweak would be "New files created in this directory belong to group opendata, regardless of the file creator's primary group" This clarifies that it is the file creators primary group and not the folder creators primary group.

PR #379, opened 2026-06-02
feat: bump FastAPI to 0.136.3 and Starlette to 1.2.1

mkitti:
## Bump FastAPI to 0.136.3 and Starlette to 1.2.1

Follow-up to #378 (stacked on `sync-dep-bounds`; this PR will retarget to `main` once #378 merges). #378 deferred the FastAPI/Starlette upgrade to keep the bounds-sync change behavior-free — this PR is that upgrade.

### Why
FastAPI was pinned `>=0.119.1,<0.120`, and `fastapi-core 0.119.1` capped Starlette at `<0.49.0`, so Starlette was stuck at 0.48.0 even though conda-forge has it through 1.2.1. Newer FastAPI (`fastapi-core 0.136.3`) **drops that cap** (`starlette >=0.46.0`, no upper bound), which unblocks Starlette 1.x.

### Changes
- `fastapi`: `>=0.119.1,<0.120` → **`>=0.136.3,<0.137`** in both the PyPI and pixi blocks.
- `starlette`: `0.48.0` → **`1.2.1`** (via `pixi update starlette`; pulled in transitively, not declared directly).
- Added **`httpx2 >=2.2.0,<3`** as a **test** dependency (see httpx note below).
- Removed the redundant explicit **`httpx`** runtime dependency (unused by our code; still provided transitively by the `fastapi` standard meta-package).
- `pixi.lock` regenerated.

### FastAPI 0.120 → 0.136 breaking-change review
Reviewed every flagged breaking change in the range against this codebase:

| Change | Version | Impact here |
|---|---|---|
| Security classes raise 401 not 403 | 0.122 | None — no `fastapi.security` usage |
| Drop Python 3.8 / 3.9 | 0.125 / 0.129 | None — we require ≥3.12 |
| Drop `pydantic.v1` | 0.126–0.128 | None — Pydantic v2 |
| Deprecate `ORJSONResponse`/`UJSONResponse` | 0.131 | None — not used |
| **`strict_content_type` (default on)** | **0.132** | **Applies** — see below |
| Underscore headers rejected | 0.136.3 | None — no `Header` params |
| SSE field validation | 0.136.2 | None — no event-stream endpoints |
| **Starlette → 1.0** | **0.136.0** | **Applies** — middleware/StaticFiles stack |

The two that touch us — `strict_content_type` and the Starlette 1.0 major bump — are both validated by the full suite (318 passing). For `strict_content_type` specifically, added a regression test (`test_set_preference_requires_json_content_type`) asserting that a JSON body sent without `application/json` is now rejected (mapped to 400 by the app's `validation_exception_handler`) rather than silently parsed.

### httpx → httpx2 note
Starlette 1.x deprecates using `httpx` with its `TestClient` and prefers `httpx2`. fileglancer has **no direct httpx usage** — it's consumed only by the TestClient at test time — so this is not a code migration. Adding `httpx2` as a test dependency makes the TestClient prefer it and silences the `StarletteDeprecationWarning`. The runtime `httpx` pin was redundant (FastAPI's standard meta-package guarantees httpx anyway), so it was removed.

### Verification
`pixi run -e test pytest` → **318 passed, no warnings** with `fastapi 0.136.3` + `starlette 1.2.1` + `httpx2 2.3.0` on Python 3.14.

### DIFF
--- a/fileglancer/filestore.py
+++ b/fileglancer/filestore.py
@@ -330,11 +330,19 @@ class Filestore:
                     full_path,
                 )
 
-        full_real = os.path.realpath(full_path)
-        if full_real == root_real:
+        # Use the resolved parent plus the basename rather than realpath(full_path)
+        # so a symlink is reported at its own location relative to root, not its
+        # target. Resolving the final component of a broken symlink follows it to a
+        # nonexistent target; on Windows an absolute target like "/nonexistent/path"
+        # resolves onto another drive and os.path.relpath then raises ValueError.
+        if full_path == root_real:
             rel_path = '.'
         else:
-            rel_path = os.path.relpath(full_real, root_real)
+            full_real = os.path.join(parent_real, os.path.basename(full_path))
+            if full_real == root_real:
+                rel_path = '.'
+            else:
+                rel_path = os.path.relpath(full_real, root_real)
 
         # Perform all filesystem stat calls here, after validation.
         lstat_result = os.lstat(full_path)
@@ -382,7 +390,14 @@ class Filestore:
         else:
             stat_result = lstat_result
 
-        full_real = os.path.realpath(full_path)
+        # Resolve the parent directory but not the entry itself, so a symlink is
+        # reported at its own location relative to root rather than at its target.
+        # Following the final component would resolve a broken symlink to its
+        # (possibly nonexistent) target; on Windows an absolute target like
+        # "/nonexistent/path" resolves onto another drive and os.path.relpath
+        # then raises ValueError ("path is on mount 'D:', start on mount 'C:'").
+        parent_real = os.path.realpath(os.path.dirname(full_path))
+        full_real = os.path.join(parent_real, entry.name)
         if full_real == root_real:
             rel_path = '.'
         else:
--- a/pixi.lock
+++ b/pixi.lock
@@ -6735,7 +6735,7 @@ packages:
 - pypi: ./
   name: fileglancer
   version: 2.8.2a0
-  sha256: eeefae4fe991f02a79f43698e2906d6053cb8e2c1dd778d2bb4378ac8a65692d
+  sha256: bae9825daa1effc319e8af45d051e8ce11b95535e2d0c829a2d892087dca9aee
   requires_dist:
   - alembic>=1.17.0,<2
   - atlassian-python-api>=4.0.7,<5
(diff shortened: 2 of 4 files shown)


mkitti:
### How this compares to #377

Both this PR and #377 land at the same resolved versions in the dev env (`fastapi 0.136.3`, `starlette 1.2.1`, `httpx2`), so they're not opposed in *outcome* — but they get there differently, and the difference matters for review and for preventing the drift from recurring.

**1. Capping vs. uncapping.** #377 reaches 0.136.3 *implicitly*: it removes the upper caps from `[tool.pixi.dependencies]` (e.g. `fastapi >=0.119.1`, uncapped) so the solver is free to float the dev lock up to whatever's current. This unblocks Starlette as a side effect, but it also re-opens the door to silent major bumps — the next solve can move `fastapi`/`starlette`/etc. again with no PR. This PR (stacked on #378) does the opposite: it pins `fastapi >=0.136.3,<0.137` *explicitly* on both surfaces, so the upgrade is a deliberate, reviewed change and the bound still protects against the next breaking minor.

**2. httpx2 as runtime vs. test dep.** #377 swaps `httpx >=0.28.1,<0.29` → `httpx2 >=2.2.0,<3` in the **runtime** `dependencies`. fileglancer has no direct httpx/httpx2 usage in application code — it's consumed only by Starlette's `TestClient` at test time, and `fastapi`'s standard meta-package already guarantees an httpx at runtime. So this PR instead adds `httpx2` as a **test** dependency and *removes* the runtime httpx pin entirely, which is where the dependency actually belongs.

**3. Verification of the behavioral changes.** Jumping 0.119 → 0.136 crosses two changes that touch this codebase — FastAPI 0.132's `strict_content_type` (JSON bodies without an `application/json` Content-Type are now rejected) and the Starlette 1.0 major bump. This PR includes a breaking-change review of the full 0.120 → 0.136 range against the code (see the PR description) and adds a regression test (`test_set_preference_requires_json_content_type`) covering the strict-Content-Type behavior. #377 doesn't add coverage for either.

**4. Reviewability.** #377 bundles cap-loosening + the httpx→httpx2 swap + the implicit fastapi/starlette jump into one PR with a ~9k-line lock diff. The #378 → #379 split separates the behavior-free bounds sync (#378) from the isolated, reviewed FastAPI/Starlette upgrade (this PR), so each can be evaluated on its own.

In short: same destination, but this PR makes the upgrade explicit and pinned, keeps `httpx2` scoped to tests, and verifies the two behavioral changes the version jump introduces.

PR #380, opened 2026-06-04
Pyright-driven cleanup

krokicki:
In order to review the changes on the `apps-improvements` branch in my IDE, I wanted things to pass error-check in Pyright. This turned into a large refactor, including a new cross-platform locking layer that cleans up some of the code while expanding its capabilities. 

With the new file locking layer, we don't need to do stuff like this:
```
try:
    import fcntl
except ImportError:
    fcntl = None  # type: ignore[assignment]
```

On the other hand, there is a lot of churn.

> **Base branch:** this PR targets `apps-improvements`, not `main`. The diff is just the cross-platform / Pyright work.

## Summary

- Add a cross-platform compatibility layer (`fileglancer/platform_compat.py`) for optional POSIX-only behavior and portalocker-backed file locking.
- Add Pyright support via `pyrightconfig.json` and a `pixi run pyright-check` task, configured with `pythonPlatform: "All"` so Linux/macOS/Windows are all validated in one run.
- Make database models, app adapters, worker IPC/S3 paths, Jira handling, and server helpers Pyright-clean across platforms.

## The cross-platform compatibility layer

`platform_compat.py` is the single place that touches platform-only modules (`pwd`, `grp`, `fcntl`/`portalocker`, Windows `ctypes`). It exposes:

- **`FileLock` / `FileLockUnavailable`** — a non-blocking exclusive file lock backed by `portalocker`, replacing the raw `fcntl.flock` calls in the job poll loop. Same semantics (one worker wins the poll election per interval), no more `try: import fcntl` guards.
- **User / group helpers** — `effective_user_home`, `named_user_home`, `user_home`, `username_for_uid`, `groupname_for_gid`, `group_names_for_user`, `current_uid`/`effective_uid`/`current_gid`, `group_ids_for_user`. These wrap the `pwd`/`grp`/`os.get*id` calls and degrade gracefully on non-POSIX.
- **`process_is_alive(pid)`** — a portable process-liveness probe (see the bug fix below).

`filestore.py`, `sshkeys.py`, `user_worker.py`, `worker_pool.py`, and `apps/core.py` were updated to route through these helpers instead of importing `pwd`/`grp`/`fcntl` directly.

## Behavioral fix: local-executor job liveness on Windows

This is the one change that is **not** just typing. `_poll_local_jobs` used `os.kill(pid, 0)` as a liveness probe. On Windows `os.kill` routes through `TerminateProcess`, so it would *kill* the job it was polling, and it doesn't raise `ProcessLookupError` for a dead PID. `process_is_alive()` now keeps the `os.kill(pid, 0)` probe on POSIX and uses a non-destructive `OpenProcess(SYNCHRONIZE)` + `WaitForSingleObject` check on Windows.

This bug was dormant because `tests/test_poll.py` was skipped wholesale on Windows via a module-level `pytest.importorskip("fcntl")`. Removing the `fcntl` dependency (now `FileLock`) un-skipped the file and surfaced it.

## SQLAlchemy 2.0 typed ORM

`database.py` migrates the models from the legacy `declarative_base()` + `Column(...)` style to the 2.0 typed style: a `DeclarativeBase` subclass with `Mapped[...]` annotations and `mapped_column(...)`. This is what makes the models type-check cleanly; column nullability is now reflected in the types (`Mapped[Optional[str]]` etc.).

## Smaller semantic changes bundled in

These came out of making the code provably correct for the type checker, and are slight behavior changes worth a look during review:

- **auth.py** — revoke the session and return `None` when `session_secret_key` is unset, instead of hashing `None`.
- **issues.py** — skip JIRA comments that lack `created`/`updated` rather than passing `None` into the required `datetime` fields (the old path could raise at model validation).
- **server.py** — add `400` validation for the JIRA ticket endpoint body; normalize proxied-path URLs with `rstrip("/")` + `HttpUrl` to avoid double slashes; rename the second `get_external_buckets` handler to `get_external_buckets_by_fsp` (resolves a shadowed function name); tighten a few response-model return annotations.
- **settings.py** — only derive `jira_browse_url` when `atlassian_url` is set.

## Tests

- `tests/test_poll.py` — dropped the module-level `importorskip("fcntl")`, reworked the lock tests onto the shared `FileLock` abstraction, and added a `spawn` fallback where `fork` is unavailable, so the poll suite runs on Windows.
- `tests/test_filestore.py` — `TestGetUserGroups` now patches `platform_compat.optional_module` instead of the removed `filestore.pwd`/`filestore.grp` globals (same assertions).
- `tests/test_worker.py` — uses `platform_compat.current_uid`.

### DIFF
--- a/fileglancer/apps/pixi.py
+++ b/fileglancer/apps/pixi.py
@@ -6,6 +6,7 @@ converting pixi tasks into runnables.
 
 import subprocess
 from pathlib import Path
+from typing import Any
 
 try:
     import tomllib
@@ -16,13 +17,14 @@ from fileglancer.model import (
     AppEntryPoint,
     AppManifest,
     AppParameter,
+    AppParameterItem,
 )
 
 _PIXI_TOML = "pixi.toml"
 _PYPROJECT_TOML = "pyproject.toml"
 
 
-def _read_pixi_config(directory: Path) -> dict | None:
+def _read_pixi_config(directory: Path) -> dict[str, Any] | None:
     """Read pixi configuration from pixi.toml or pyproject.toml.
 
     Returns the pixi config dict (equivalent to the [tool.pixi] section),
@@ -41,15 +43,18 @@ def _read_pixi_config(directory: Path) -> dict | None:
     pyproject_path = directory / _PYPROJECT_TOML
     if pyproject_path.is_file():
         data = tomllib.loads(pyproject_path.read_text())
-        pixi_config = data.get("tool", {}).get("pixi")
-        if pixi_config and "tasks" in pixi_config:
+        tool_config = data.get("tool", {})
+        pixi_config = tool_config.get("pixi") if isinstance(tool_config, dict) else None
+        if isinstance(pixi_config, dict) and "tasks" in pixi_config:
             # Merge top-level [project] metadata into pixi config.
             # [tool.pixi.project] may only have channels/platforms,
             # so fill in missing fields (name, description, version)
             # from the top-level [project] section.
             top_project = data.get("project", {})
-            if top_project:
+            if isinstance(top_project, dict) and top_project:
                 pixi_project = pixi_config.get("project", {})
+                if not isinstance(pixi_project, dict):
+                    pixi_project = {}
                 for key in ("name", "description", "version"):
                     if key not in pixi_project and key in top_project:
                         pixi_project[key] = top_project[key]
@@ -60,27 +65,37 @@ def _read_pixi_config(directory: Path) -> dict | None:
     return None
 
 
-def _collect_tasks(config: dict) -> dict[str, dict]:
+def _collect_tasks(config: dict[str, Any]) -> dict[str, dict[str, Any]]:
     """Collect all tasks from a pixi config, including feature tasks.
 
     Returns a dict of task_name -> task_definition (normalized to dict form).
     """
-    tasks: dict[str, dict] = {}
+    tasks: dict[str, dict[str, Any]] = {}
 
     # Top-level tasks
-    for name, defn in config.get("tasks", {}).items():
-        tasks[name] = _normalize_task(defn)
+    task_config = config.get("tasks", {})
+    if isinstance(task_config, dict):
+        for name, defn in task_config.items():
+            tasks[str(name)] = _normalize_task(defn)
 
     # Feature tasks (e.g. [tool.pixi.feature.test.tasks])
-    for feature_name, feature in config.get("feature", {}).items():
-        for name, defn in feature.get("tasks", {}).items():
-            if name not in tasks:
-                tasks[name] = _normalize_task(defn)
+    feature_config = config.get("feature", {})
+    if isinstance(feature_config, dict):
+        for feature in feature_config.values():
+            if not isinstance(feature, dict):
+                continue
+            feature_tasks = feature.get("tasks", {})
+            if not isinstance(feature_tasks, dict):
+                continue
+            for name, defn in feature_tasks.items():
+                task_name = str(name)
+                if task_name not in tasks:
+                    tasks[task_name] = _normalize_task(defn)
 
     return tasks
 
 
-def _normalize_task(defn) -> dict:
+def _normalize_task(defn: Any) -> dict[str, Any]:
     """Normalize a task definition to dict form.
 
     Tasks can be defined as:
@@ -89,10 +104,12 @@ def _normalize_task(defn) -> dict:
     """
     if isinstance(defn, str):
         return {"cmd": defn}
-    return dict(defn)
+    if isinstance(defn, dict):
+        return dict(defn)
+    return {}
 
 
-def _convert_task_arg(arg) -> AppParameter:
+def _convert_task_arg(arg: Any) -> AppParameter:
     """Convert a pixi task argument to an AppParameter."""
     if isinstance(arg, str):
         # Simple required argument
@@ -103,7 +120,7 @@ def _convert_task_arg(arg) -> AppParameter:
         )
 
     # Dict form with optional default and choices
-    arg_name = arg["arg"]
+    arg_name = str(arg["arg"])
     param_type = "enum" if "choices" in arg else "string"
     kwargs: dict = {
         "name": arg_name.replace("_", " ").title(),
@@ -118,7 +135,7 @@ def _convert_task_arg(arg) -> AppParameter:
     return AppParameter(**kwargs)
 
 
-def _task_to_entry_point(name: str, task: dict) -> AppEntryPoint | None:
+def _task_to_entry_point(name: str, task: dict[str, Any]) -> AppEntryPoint | None:
     """Convert a pixi task dict to an AppEntryPoint.
 
     Returns None for tasks without a command (dependency-only tasks with no
@@ -144,7 +161,7 @@ def _task_to_entry_point(name: str, task: dict) -> AppEntryPoint | None:
     description = task.get("description")
 
     # Convert task arguments to parameters
-    parameters = []
+    parameters: list[AppParameterItem] = []
     args = task.get("args")
     if args:
         for arg in args:
@@ -211,9 +228,12 @@ class PixiAdapter:
 
     def convert(self, directory: Path) -> AppManifest:
         config = _read_pixi_config(directory)
+        if config is None:
+            raise ValueError("No pixi configuration found")
 
         # Extract project metadata
-        project = config.get("project", config.get("workspace", {}))
+        project_config = config.get("project") or config.get("workspace") or {}
+        project = project_config if isinstance(project_config, dict) else {}
         description = project.get("description")
         version = project.get("version")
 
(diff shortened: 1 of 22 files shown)


krokicki:
@mkitti I'd appreciate your thoughts on this one.

fileglancer/apps/pixi.py
```diff
@@ -60,27 +65,37 @@ def _read_pixi_config(directory: Path) -> dict | None:
...
+            for name, defn in feature_tasks.items():
+                task_name = str(name)
+                if task_name not in tasks:
+                    tasks[task_name] = _normalize_task(defn)
 
     return tasks
 
 
-def _normalize_task(defn) -> dict:
+def _normalize_task(defn: Any) -> dict[str, Any]:
```
mkitti, line 98: There should be some point where the dict definition becomes more strict. Could we constrain the value type here?

mkitti (commented):
If we are going to use typing, `dict[str, Any]` is probably something we want to move away from when possible. It would be be better to create a dataclass with an `asdict()` method. You could also consider using `collections.abc.Mapping`.

fileglancer/apps/pixi.py
```diff
@@ -89,10 +104,12 @@ def _normalize_task(defn) -> dict:
...
     if isinstance(defn, str):
         return {"cmd": defn}
-    return dict(defn)
+    if isinstance(defn, dict):
+        return dict(defn)
+    return {}
 
 
-def _convert_task_arg(arg) -> AppParameter:
+def _convert_task_arg(arg: Any) -> AppParameter:
```
mkitti, line 112:
```suggestion
def _convert_task_arg(arg: str | Dict[str, Any]) -> AppParameter:
```

mkitti:
@d-v-b suggested looking into `TypedDict`: https://typing.python.org/en/latest/spec/typeddict.html

fileglancer/database.py
```diff
@@ -495,7 +501,7 @@ def get_all_user_preferences(session: Session, username: str) -> Dict[str, Dict]
     return {pref.key: pref.value for pref in prefs}
 
 
-def get_proxied_paths(session: Session, username: str, fsp_name: str = None, path: str = None) -> List[ProxiedPathDB]:
+def get_proxied_paths(session: Session, username: str, fsp_name: Optional[str] = None, path: Optional[str] = None) -> List[ProxiedPathDB]:
```
mkitti, line 504:
```suggestion
def get_proxied_paths(session: Session, username: str, fsp_name: str | None = None, path: str | None = None) -> List[ProxiedPathDB]:
```

https://typing.python.org/en/latest/guides/modernizing.html

> While Union and Optional are not considered obsolete, using the | (pipe) operator is often more readable.

d-v-b:
please use typeddicts when you know something about the keys / values of your dicts 🙏

fileglancer/database.py
```diff
@@ -818,7 +824,7 @@ def delete_neuroglancer_state(session: Session, username: str, short_key: str) -
     return deleted
 
 
-def get_tickets(session: Session, username: str, fsp_name: str = None, path: str = None) -> List[TicketDB]:
+def get_tickets(session: Session, username: str, fsp_name: Optional[str] = None, path: Optional[str] = None) -> List[TicketDB]:
```
mkitti, line 827:
```suggestion
def get_tickets(session: Session, username: str, fsp_name: str | None = None, path: str | None = None) -> List[TicketDB]:
```

fileglancer/filestore.py
```diff
@@ -134,7 +133,7 @@ def _get_symlink_target_fsp(cls, absolute_path: str, is_symlink: bool,
     @classmethod
     def from_stat(cls, path: str, absolute_path: str,
                   lstat_result: os.stat_result, stat_result: os.stat_result,
-                  current_user: str = None, fsps: Optional[list] = None,
+                  current_user: Optional[str] = None, fsps: Optional[list] = None,
```
mkitti, line 136:
```suggestion
                  current_user: str | None] = None, fsps: list | None = None,
```

fileglancer/filestore.py
```diff
@@ -134,7 +133,7 @@ def _get_symlink_target_fsp(cls, absolute_path: str, is_symlink: bool,
...
                   root_path: Optional[str] = None,
                   user_groups: Optional[set[str]] = None):
```
mkitti, line 138:
```suggestion
                  root_path: str | None] = None,
                  user_groups: set[str] | None = None):
```

mkitti, line 138:
```suggestion
                   root_path: str | None] = None,
                   user_groups: set[str] | None = None):
```

fileglancer/filestore.py
```diff
@@ -281,7 +260,7 @@ def _check_path_in_root(self, path: Optional[str]) -> str:
         return full_path
 
 
-    def _get_file_info_from_path(self, full_path: str, current_user: str = None,
+    def _get_file_info_from_path(self, full_path: str, current_user: Optional[str] = None,
```
mkitti, line 263:
```suggestion
    def _get_file_info_from_path(self, full_path: str, current_user: str | None = None,
```

fileglancer/filestore.py
```diff
@@ -356,7 +335,7 @@ def _is_within_root(p: str) -> bool:
         )
 
 
-    def _file_info_from_direntry(self, entry: os.DirEntry, current_user: str = None,
+    def _file_info_from_direntry(self, entry: os.DirEntry, current_user: Optional[str] = None,
```
mkitti, line 338:
```suggestion
    def _file_info_from_direntry(self, entry: os.DirEntry, current_user: str | None = None,
```

fileglancer/filestore.py
```diff
@@ -433,7 +412,7 @@ def get_absolute_path(self, relative_path: Optional[str] = None) -> str:
         return os.path.abspath(os.path.join(self.root_path, relative_path))
 
 
-    def get_file_info(self, path: Optional[str] = None, current_user: str = None,
+    def get_file_info(self, path: Optional[str] = None, current_user: Optional[str] = None,
                       fsps: Optional[list] = None) -> FileInfo:
```
mkitti, line 416:
```suggestion
                       fsps: list | None = None) -> FileInfo:
```

fileglancer/filestore.py
```diff
@@ -492,7 +471,7 @@ def check_is_binary(self, path: Optional[str] = None, sample_size: int = 4096) -
             return True
 
 
-    def yield_file_infos_paginated(self, path: Optional[str] = None, current_user: str = None,
+    def yield_file_infos_paginated(self, path: Optional[str] = None, current_user: Optional[str] = None,
```
mkitti, line 474:
```suggestion
    def yield_file_infos_paginated(self, path: sre | None = None, current_user: str | None = None,
```

fileglancer/filestore.py
```diff
@@ -569,7 +548,7 @@ def yield_file_infos_paginated(self, path: Optional[str] = None, current_user: s
         next_cursor = page_entries[-1].name if has_more and page_entries else None
         return file_infos, has_more, next_cursor, total_count, is_truncated
 
-    def yield_file_infos(self, path: Optional[str] = None, current_user: str = None,
+    def yield_file_infos(self, path: Optional[str] = None, current_user: Optional[str] = None,
```
mkitti, line 551:
```suggestion
    def yield_file_infos(self, path: str | None = None, current_user: str | None = None,
```

fileglancer/filestore.py
```diff
@@ -604,7 +583,7 @@ def yield_file_infos(self, path: Optional[str] = None, current_user: str = None,
                 continue
 
 
-    def stream_file_contents(self, path: str = None, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
+    def stream_file_contents(self, path: Optional[str] = None, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
```
mkitti, line 586:
```suggestion
    def stream_file_contents(self, path: str | None = None, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
```

fileglancer/filestore.py
```diff
@@ -640,7 +619,7 @@ def stream_file_contents(self, path: str = None, buffer_size: int = DEFAULT_BUFF
                         break
                     yield chunk
 
-    def stream_file_range(self, path: str = None, start: int = 0, end: int = 0, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
+    def stream_file_range(self, path: Optional[str] = None, start: int = 0, end: int = 0, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
```
mkitti, line 622:
```suggestion
    def stream_file_range(self, path: str | None = None, start: int = 0, end: int = 0, buffer_size: int = DEFAULT_BUFFER_SIZE, file_handle = None) -> Generator[bytes, None, None]:
```

mkitti (changes requested):
Use `| None` rather than `Optional` per https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/consider-alternative-union-syntax.html

Issue #386, opened 2026-06-11
Nextflow apps cannot be created if the branch is not main


cgoina:
I tried to create an app for the EASI-FISH nextflow pipeline using the ome-zarr-dev branch but the app could not be created. The error was that the runnables.yaml was missing but the branch has a nextflow_schema.json. When I changed the branch from ome-zarr-dev to main the app was created

<img width="506" height="372" alt="Image" src="https://github.com/user-attachments/assets/0ca326b2-1029-4a06-a94a-22be3332d667" />

allison-truhlar:
Closed with merging #387

PR #387, opened 2026-06-11
Fix for options that can have multiple values other than strings

cgoina:
This is the fix for https://github.com/JaneliaSciComp/fileglancer/issues/386. The problem was not related to the branch as the ticket mentioned but the schema optional values. The branch failed because of this particular parameter which was only in the branch and not in main:
```
                "segmentation_zarr_format": {
                    "type": "integer",
                    "default": 2,
                    "enum": [2, 3],
                    "description": "Zarr format version (2 or 3) for segmentation output zarr containers",
                    "fa_icon": "fas fa-file-archive"
                },
```

### DIFF
--- a/fileglancer/apps/core.py
+++ b/fileglancer/apps/core.py
@@ -270,7 +270,8 @@ def _find_manifests_in_repo(repo_dir: Path) -> list[tuple[str, AppManifest]]:
             if adapter.can_handle(repo_dir):
                 results.append(("", adapter.convert(repo_dir)))
         except Exception as e:
-            logger.warning(f"Adapter {type(adapter).__name__} failed: {e}")
+            raise ValueError(f"Adapter {type(adapter).__name__} failed: {e}")
+
 
     return results
 
--- a/fileglancer/model.py
+++ b/fileglancer/model.py
@@ -334,7 +334,7 @@ class AppParameter(BaseModel):
     description: Optional[str] = Field(description="Description of the parameter", default=None)
     required: bool = Field(description="Whether the parameter is required", default=False)
     default: Optional[Any] = Field(description="Default value for the parameter", default=None)
-    options: Optional[List[Any]] = Field(description="Allowed values for enum type", default=None)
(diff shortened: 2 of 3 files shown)


fileglancer/apps/core.py
```diff
@@ -270,7 +270,8 @@ def _find_manifests_in_repo(repo_dir: Path) -> list[tuple[str, AppManifest]]:
             if adapter.can_handle(repo_dir):
                 results.append(("", adapter.convert(repo_dir)))
         except Exception as e:
-            logger.warning(f"Adapter {type(adapter).__name__} failed: {e}")
+            raise ValueError(f"Adapter {type(adapter).__name__} failed: {e}")
+
```
cgoina, line 274: Raising an exception here displays the actual error in the UI

fileglancer/model.py
```diff
@@ -334,7 +334,7 @@ class AppParameter(BaseModel):
     description: Optional[str] = Field(description="Description of the parameter", default=None)
     required: bool = Field(description="Whether the parameter is required", default=False)
     default: Optional[Any] = Field(description="Default value for the parameter", default=None)
-    options: Optional[List[str]] = Field(description="Allowed values for enum type", default=None)
+    options: Optional[List[Any]] = Field(description="Allowed values for enum type", default=None)
```
neomorphic: Is `Any` too broad here? Can it be narrowed down?

cgoina: it can - so far I only use str|int but I can see somebody may use float so we can make it List[str|int|float]

neomorphic (approved):
LGTM

fileglancer/apps/core.py
krokicki: In the future we'll probably want to capture all these parse errors (and also the the JSON parse exceptions on line 257) and present them to the user, but keep processing, so that one bad manifest doesn't break the entire repo. But I think this is reasonable for now.

krokicki (approved):
LGTM

allison-truhlar (approved):
Works for me, thanks Cristian!

PR #389, opened 2026-06-17
App Catalog 

krokicki:
This PR adds an **App Catalog** that lets users publish ("share") apps they've added to a shared, server-wide catalog, and lets other users add those listings to their own app collection. It also reworks the Apps area into a tabbed layout, moves tool requirement verification from submit-time (server) to job runtime (compute node), and removes the confusing manifest `version` field.

### Backend — catalog

- **New table `app_listings`** (`fileglancer/database.py`, migration `d9f1a3c5e208_add_app_listings_table.py`): `owner_username`, `url`, `manifest_path`, `branch`, `name`, `description`, `published_at`, `updated_at`, with a unique constraint on `(owner_username, url, manifest_path)` and an index on `owner_username`.
- **CRUD helpers** in `database.py`: `list_app_listings`, `get_app_listing`, `get_app_listings_by_owner`, `get_app_listing_for_app`, `create_app_listing` (rejects duplicates), `update_app_listing` (owner-scoped, editable name/description), `delete_app_listing` (owner-scoped).
- **New models** in `model.py`: `AppListing`, `ShareAppRequest`, `UpdateAppListingRequest`, plus `validate_catalog_listing_name` / `resolve_catalog_listing_name` (non-empty, trimmed). `UserApp` gains an optional `listing_id` so the UI knows whether the user has already shared it.
- **New endpoints** in `server.py`:
  - `GET /api/catalog` — list all listings (newest first)
  - `POST /api/catalog` — share one of the caller's own apps (404 if not owned, 409 on duplicate)
  - `PATCH /api/catalog/{id}` — edit name/description of a listing you own
  - `DELETE /api/catalog/{id}` — unshare a listing you own
  - `POST /api/catalog/{id}/add` — re-fetch the manifest/branch and upsert the listing's app into the caller's own apps (409 if already added)
  - `GET /api/user/apps` now annotates each app with its `listing_id`.

### Backend — requirements at runtime instead of submit-time

- `submit_job` no longer calls `verify_requirements` (server PATH). Instead, `build_requirements_check()` (`apps/core.py`) emits a bash snippet injected into the job script after PATH/conda/env setup but before `pre_run`/command. On any unmet requirement it prints all errors to stderr and `exit 1`s, failing the job with a readable message.
- Requirement parsing tightened to **reject compound/chained/comma specs** (`pixi>=0.40,<0.60`), shared via `_validate_requirements` in `model.py` (used by both `AppManifest` and `AppEntryPoint`).

### Removed: manifest `version` field

- Dropped from `AppManifest` (`model.py`), the pixi/nextflow adapters, the TS `AppManifest` type, `AppInfoDialog`, `AppLaunchForm`, and `AuthoringApps.md`.

### Frontend — tabbed Apps area

- New `AppsLayout` with tabs **My Apps** / **App Catalog** / **Jobs** (job-count badge), nested routes under `/apps` (`App.tsx`). Navbar collapses the separate "Jobs" item into "Apps".
- New `Catalog` page with search (name/description/sharer) and a "hide already installed" filter; `ListingCard` + `ListingInfoDialog`.
- New `DeleteAppDialog` confirmation before removing an app.
- `AppInfoDialog` gains an inline "Share to Catalog" form (keeps the dialog open through sharing) and Share/Unshare actions; `AppCard` shows a "Shared" badge.
- New TanStack Query hooks in `appsQueries.ts`: `useCatalogQuery`, `useShareAppMutation`, `useUpdateListingMutation`, `useUnshareListingMutation`, `useAddFromListingMutation`.
- `FgButton` loading state now overlays the spinner without changing button size; `Spinner` accepts `sizeClasses` and renders no text node when `text` is unset.

### Tests

- `tests/test_apps.py`: `build_requirements_check` coverage (present/missing/version compare/compound-reject/pipefail).
- `tests/test_catalog_endpoints.py`: new endpoint coverage (~417 lines).

## Incidental fixes to pre-existing issues

These were not introduced by the catalog work — they pre-dated it or arrived via other changes bundled on the branch — and were cleaned up along the way:

- **`_find_manifests_in_repo` raised on the first adapter failure** (`apps/core.py`), so one adapter's error could mask a later adapter that would have handled the repo. This behavior came from the ticket-386 "multiple values" change (commit `1592067c`), not the catalog work. Fixed to collect all adapter errors, return as soon as any adapter succeeds, and raise an aggregated error only when no adapter produced a manifest. Tests added.
- **`merge_requirements` matched the same requirement regex up to three times per entry** (`apps/core.py`). Pre-existing on `main`; tidied to match once while consolidating the requirement-spec regex.


krokicki:
I also added import/export of parameters here, since that is critical for running Nextflow pipelines. Then I optimized the performance of the jobs pages.

PR #395, opened 2026-06-23
Grant Private Network Access preflights for /files/ data links

krokicki:
## Problem

Browsers are tightening access from public-origin pages to private-network hosts. Chromium (Chrome/Edge) does this via **Private Network Access (PNA)**: before a request from a public-origin page (e.g. `https://neuroglancer-demo.appspot.com`) to a private-network address (e.g. an internal host serving Fileglancer's `/files/` data links), the browser sends a CORS preflight carrying `Access-Control-Request-Private-Network: true`. The request only proceeds if the response echoes `Access-Control-Allow-Private-Network: true`.

Starlette's `CORSMiddleware` does not emit that header, so Chromium blocks cross-origin viewers (Neuroglancer/N5/Vizarr) from loading data hosted on an internal network.

## Change

Add a small pure-ASGI `PrivateNetworkAccessMiddleware` (modeled on the existing `RequestIdMiddleware`) that appends `Access-Control-Allow-Private-Network: true` to responses for preflights carrying the PNA request header. It is registered **outside** `CORSMiddleware` so it wraps the preflight response that CORS generates. The header is added only when the PNA request header is present (sent by browsers solely on preflights), so it never leaks onto normal data responses.

## Tests

- `test_pna_preflight_grants_private_network` — preflight with the PNA request header gets the grant header back.
- `test_pna_header_absent_without_request` — the grant header does not appear on ordinary responses.

## Note on Firefox

This does **not** address Firefox, which uses a user-permission model (Local Network Access) rather than a server header. The Firefox remedy is client/policy-side (`network.lna.skip-domains`, or the enterprise `LocalNetworkAccess` policy). This PR gets ahead of the equivalent Chromium breakage.

A mirroring change [is in x2s3](https://github.com/JaneliaSciComp/x2s3/pull/21), since the two share this CORS surface.

### DIFF
--- a/fileglancer/server.py
+++ b/fileglancer/server.py
@@ -87,6 +87,56 @@ class RequestIdMiddleware:
         await self.app(scope, receive, send_with_request_id)
 
 
+class PrivateNetworkAccessMiddleware:
+    """Pure ASGI middleware that grants browser Private Network Access (PNA)
+    preflights.
+
+    Chromium browsers (Chrome/Edge) send a CORS preflight before any request from
+    a public-origin page (e.g. https://neuroglancer-demo.appspot.com) to a
+    private-network address (e.g. an internal host serving Fileglancer's /files/
+    data links). The preflight carries `Access-Control-Request-Private-Network: true`,
+    and the request only proceeds if the response echoes
+    `Access-Control-Allow-Private-Network: true`. Starlette's CORSMiddleware does not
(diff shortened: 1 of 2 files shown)


neomorphic (approved):
Looks good to me. Maybe we could turn this middleware into a package, since it is getting used in multiple locations.

krokicki:
@neomorphic That's a good idea.

PR #396, opened 2026-06-23
Run the devcontainer under rootless Podman

krokicki:
Makes the devcontainer runnable under **rootless Podman** on Linux while staying fully compatible with Docker/Colima on Mac, and hardens the egress firewall so the AI agent (which runs with approvals disabled) cannot loosen its own network constraints.

Runtime is selected with the `FG_CONTAINER_RUNTIME` environment variable (default: `docker`). `devcontainer.json` stays runtime-neutral; all Podman-specific behavior lives in a small wrapper and shim.

## Changes

**Cross-platform runtime selection**
- `pyproject.toml`: the four `container-*` tasks now call `.devcontainer/dc.sh`.
- `.devcontainer/dc.sh` (new): passes `--docker-path` to the devcontainers CLI based on `FG_CONTAINER_RUNTIME`. On the Podman path it also routes through the shim and disables the CLI's uid-renumber step.
- `.devcontainer/podman-shim/podman` (new): upgrades the CLI's forced `--userns=keep-id` to `keep-id:uid=1000,gid=1000` so the host user maps onto the container's `vscode` user (uid 1000). This makes bind mounts writable using only a small subuid range, avoiding the unsafe ~1M-id range the CLI's renumber step would otherwise require on a shared LDAP host.

**Tamper-proof egress firewall**
- `.devcontainer/init-firewall.sh`: DNS egress is pinned to the resolvers in `/etc/resolv.conf` instead of allowing port 53 to any host; the blanket outbound-22 rule is removed (SSH to GitHub still works via the allowed-domains set, SSH to other hosts is blocked); the blanket host `/24` allow is removed.
- `.devcontainer/post-create.sh`: revokes the `vscode` user's passwordless sudo as the final setup step. Without root, the agent cannot touch iptables/ipset, so the egress allowlist can no longer be flushed from inside the container. `NET_ADMIN`/`NET_RAW` remain in the image but are inert without root. A root shell is still available from the host via `podman exec -u root` / `docker exec -u root`.

## Verification

Rebuilt under rootless Podman and confirmed on the running container:
- `vscode` can no longer `sudo`, and both `iptables -F` and `sudo iptables -F` fail.
- Egress policy shows DNS pinned to the three configured resolvers, no port-22-anywhere, no `/24`.
- Normal egress still works: `api.github.com` returns 200, `example.com` is blocked.
- Bind-mounted credentials are owned by `vscode` and readable; Claude and Codex run.
- Host-side `podman exec -u root` still works for maintenance.

## Notes

- No changes to `devcontainer.json`; Mac/Colima continues to use the default Docker path and never sees the shim.
- This bounds blast radius and prevents the agent from removing its own egress controls. It does not stop exfiltration through already-allowed endpoints (e.g. GitHub, the Anthropic/OpenAI APIs) or DNS tunneling through the permitted resolver; the structural fix for that is enforcing egress outside the container (e.g. a VM/hypervisor boundary).


mkitti:
Is rootless podman meant for usage similar to how we might use apptainer on HPC?

krokicki:
This is meant to replace our use of Docker entirely, especially on Linux. Running agents inside of Docker is not safe because it effectively gives them root access to the system. On Mac we have other good options like Colima (though Podman also works there, for those who prefer it).

mkitti:
Yes, I understand why we are doing this. Those same attributes make it seem amenable to running as an unprivileged user on a cluster as well.

PR #410, opened 2026-07-19
Apps UX improvements

krokicki:
A set of small, independent quality-of-life improvements to the Apps feature — catalog, launch flow, and the app/listing/job detail pages. Each ships as its own commit; the four core items are captured in `docs/superpowers/specs/2026-07-18-apps-improvements-design.md`.

## Navigation: breadcrumbs across the Apps pages

The launch page and the app/listing detail pages now show a breadcrumb trail (apps-home icon → app → entry point, `›`-delimited) in place of the old back-arrow + title header, via a new `AppBreadcrumbs` component.

- The trail's origin is inferred from install status with no router plumbing: installed apps route home → **My Apps** and back to the app detail page; not-installed apps route home → **App Catalog** and back to the matching catalog listing (resolved from the cached catalog query by canonical URL + manifest path).
- Launch breadcrumbs point back to wherever the user came from, and catalog-originated launches keep the **App Catalog** tab selected.

## Catalog

- **Install count** — a new sortable "Installs" column and a count on the listing detail page. An install is counted live from `user_apps` (grouped `COUNT` by url + manifest path, joined in memory) — no new column, no migration, no counter to keep in sync.
- **Leaner table** — dropped the Repository column and let the Description column take the remaining width; the app name no longer includes the entry point.

## Add-app flow

Pasting a GitHub URL that embeds a branch/tag (e.g. `.../tree/my-branch`) now splits the ref into the Revision field automatically, leaving a bare repo URL in the URL field. A pasted ref is authoritative and overwrites the Revision box; typing a bare URL is undisturbed. Backed by a new pure helper `splitGithubRef`.

## Commit display

Standardized the commit SHA display: the app page now shows the short 7-char SHA (matching the job page) and both the "App commit" and job-page "Version" labels are renamed to **Commit**.

## Working directory: new `repo` option

The `working_dir` value that ran from the manifest's directory inside the clone is renamed to **`manifest`** (the new default for non-container runnables). **`repo`** is reintroduced to always run from the clone root, ignoring the manifest's subdirectory — useful for a manifest nested in a monorepo whose command must run from the top of the project. The cached clone is now bind-mounted into containers for both `manifest` and `repo` (anything but `work`).

**Not backwards compatible**: in the separate-repos case (manifest and code in different repos), `manifest` now `cd`s into `repo/<manifest_path>` for consistency with the same-repo case, where it previously ran from the clone root. `FG_MANIFEST_DIR` (added on `main`) is reconciled with the rename and likewise points at `repo[/manifest_path]`.

## Testing

- Backend: new pytest coverage for the catalog install count.
- Frontend: unit tests for `splitGithubRef` (bare URL, `/tree/<branch>`, `/tree/main`, `.git` suffix, non-GitHub string) and component tests for `AppBreadcrumbs`.

### DIFF
--- a/docs/superpowers/specs/2026-07-18-apps-improvements-design.md
+++ b/docs/superpowers/specs/2026-07-18-apps-improvements-design.md
@@ -1,0 +1,47 @@
+# Apps feature — minor improvements design (2026-07-18)
+
+Four small, independent improvements to the Apps feature. Each ships as its own commit.
+
+## 1. Paste a GitHub URL with an embedded branch/tag → move it to the Revision box
+
+When adding an app, a user often pastes a URL copied from GitHub's branch/tag view, e.g. `https://github.com/org/repo/tree/my-branch`. Today that whole string goes into the URL field and the Revision field stays empty; the ref survives only because `buildAppUrl` re-parses it later. We make the split explicit and visible.
+
+`parseGithubUrl` (`frontend/src/utils/appUrls.ts`) already extracts `{owner, repo, branch}` from a `/tree/<ref>` URL. Add a one-line pure helper `splitGithubRef(url)` that returns `{ repoUrl, ref }` — the bare `https://github.com/owner/repo` plus the embedded ref (empty string when the URL carried no ref or the ref is `main`).
+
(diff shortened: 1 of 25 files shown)


krokicki:
TODO: 
- [ ] Add tooltip for app catalog name which shows the full repo name

PR #430, opened 2026-08-15
Warn about Zarr layouts that make viewing awkward

krokicki:
Adds two advisory warnings to the Zarr preview panel, flagging layout choices that make a dataset expensive to view. They run client-side on metadata fileglancer already parses, so there are no extra requests on the common path.

Motivation is the production incident that occurred on Friday, August 14th. The IO thread pools were starved by a handful of viewing sessions on a 91 GB image published with a single resolution level. Nothing here changes server behaviour — the aim is to tell data creators before they publish.

## Screenshot

Screenshot 2026-08-21 at 1 50 31 PM

## The checks

**Only one resolution level.** A multiscales group declaring a single dataset provides no downsampled data, so every zoom reads full resolution. Keyed on the dataset count rather than the number of shapes, because a plain zarr array also has exactly one shape while claiming nothing about being multiscale — an early version keyed on shapes and fired on healthy pyramid levels that users had browsed into. Gated at 1 GB.

**Chunks too large.** Chunk size affects viewing because a viewer fetches a whole chunk to show any part of it. Sizes come from shape and dtype, so they are logical: exact for a raw array, an upper bound for a compressed one. The limit is 32 MB with no compressor and 64 MB with one, since the compression ratio is a property of the pixels and is not in the metadata. Unknown or unfetched codec info counts as compressed, so uncertainty costs a missed warning rather than a false one.

The threshold sits above every published recommendation rather than at any one of them — 1–10 MB from [the image.sc discussion](https://forum.image.sc/t/should-compression-play-a-role-in-selecting-chunk-sizes-for-ome-zarr-v0-4-datasets/117877), 8–16 MB for S3 byte-range reads in [Amazon's S3 best practices](https://d1.awsstatic.com/whitepapers/AmazonS3BestPractices.pdf), 32³–128³ voxel inner chunks from [webknossos](https://docs.webknossos.org/webknossos/data/zarr.html). A chunk past 32 MB on disk is outside the range everyone recommends, not merely on the large side of it. The 128 MB chunks from the incident still fire; 256³ uint16 does not.

Codec classification reuses `capability-manifest`'s `classifyCodec`, plus a walk over the pipeline it does not do itself: a sharded v3 array lists only the structural `sharding_indexed` codec at the top level and carries the real compressor in its configuration, so a flat scan calls every sharded array uncompressed.

Sharded arrays are measured by their inner chunks, and the warning says so. zarrita prefers the sharding codec's `chunk_shape` over `chunk_grid` when it opens an array, so `arr.chunks` is already the unit a viewer fetches and never the shard around it.

## Notes for review

- **Nothing is withheld.** Both are advisory; viewer links and thumbnails are untouched. Neuroglancer fetches only the chunks intersecting the current view, so it opens these datasets fine.
- **No axis-order check.** An earlier revision warned when axes deviated from the `t, c, z, y, x` order the spec requires. Removed: `ome-zarr.js` selects the image plane by axis *name* (`getSlices` gives `x`/`y` full slices and collapses `z`/`t`/`c`), so a non-spec order renders transposed rather than as the cross-section the warning claimed, and channel-last layouts that render correctly were being flagged. The remaining harm is a mirrored thumbnail, which is a bug on our side of the wire and not worth asking anyone to rewrite an array over. Two upstream positional assumptions came out of that reading and are worth fixing regardless: the render reads `shape[0]`/`shape[1]` as height/width after name-driven slicing, and the thumbnail size guard multiplies `shape[len-1] * shape[len-2]`, which is the wrong plane when the axes are permuted.
- **No thumbnail check.** `ome-zarr.js` already refuses when the lowest level exceeds its own limit. Modelling its behaviour a second time proved unreliable: it renders from the level nearest the requested size rather than the smallest, and reads a single plane rather than a whole level.
- `zarrQueries.ts` now populates codec info for plain arrays too — the v3 branch reads it from the `zarr.json` it already parsed, the v2 branch fetches `.zarray` as the group path does.
- `__tests__/mocks/omezarrHelper.ts` is a whole-module mock, so it needs every new export listed or the component tests fail.
- Out of scope, worth separate issues: chunk dimensionality vs array dimensionality, a minimum chunk size, adaptive display that adjusts to array properties instead of warning, and on-the-fly or materialized pyramid generation as a service.

## Testing

12 unit tests covering both compression paths, nested sharding codecs, unknown codecs, the sharded inner-chunk case, and the plain-array cases that earlier versions got wrong. Full frontend suite passes (316). Verified end to end against a pair of synthetic fixtures that reproduce the original dataset's defects and the corrected version.

@StephanPreibisch @JaneliaSciComp/fileglancer @dchen116

### DIFF
--- a/frontend/src/components/ui/BrowsePage/MetadataHint.tsx
+++ b/frontend/src/components/ui/BrowsePage/MetadataHint.tsx
@@ -14,8 +14,12 @@ type MetadataHintVariant =
   | { case: 'zarr-query-error'; errorMessage?: string }
   // Zarr - metadata is valid, but the layout will make viewing awkward
   | { case: 'zarr-single-level'; size: string }
-  | { case: 'zarr-large-chunks'; size: string; compressed: boolean }
-  | { case: 'zarr-axis-order'; axisOrder: string; expectedOrder: string }
+  | {
+      case: 'zarr-large-chunks';
+      size: string;
+      compressed: boolean;
+      sharded: boolean;
+    }
   // N5 - query never fired
   | { case: 'n5-has-s0-no-attrs' }
   | { case: 'n5-has-attrs-no-s0' }
@@ -80,19 +84,13 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
       return {
         kind: 'warning',
         title: 'Only one resolution level',
-        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Viewers must read full-resolution at every zoom level, so viewing the whole image is far more expensive than it needs to be. Generating a multiscale pyramid fixes this.`
-      };
-    case 'zarr-axis-order':
-      return {
-        kind: 'warning',
-        title: 'Axes are not in the order OME-Zarr specifies',
-        description: `The axes are ordered ${variant.axisOrder}, but the spec requires T, C, Z, Y, X order - here that would be ${variant.expectedOrder}. Many tools take the last two axes to be the image plane, so they will show a cross-section rather than the expected view. Rewriting the dataset with the axes in spec order avoids this.`
+        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Without multiple levels, viewers read the full-resolution data at every zoom level, making viewing slow. Generating a multiscale pyramid fixes this.`
       };
     case 'zarr-large-chunks':
       return {
         kind: 'warning',
         title: 'Chunks may be too large for efficient viewing',
-        description: `This dataset uses ${variant.size} chunks ${variant.compressed ? '(before compression)' : '(without compression)'}. Chunk files larger than the browser's cache limit are re-downloaded on every access, which makes viewing slow. A final chunk size of 1-10 MB works best.`
+        description: `This dataset uses ${variant.size} ${variant.sharded ? 'inner chunks' : 'chunks'} ${variant.compressed ? '(before compression)' : '(without compression)'}. Very large chunks make viewing slow, because a viewer must fetch a whole chunk to show any part of it. A stored chunk size of 1-32 MB works best.`
       };
     case 'n5-has-s0-no-attrs':
       logger.info(
--- a/frontend/src/components/ui/BrowsePage/MetadataHint.tsx
+++ b/frontend/src/components/ui/BrowsePage/MetadataHint.tsx
@@ -12,6 +12,10 @@ type MetadataHintVariant =
   | { case: 'zarr-v2-no-multiscales' }
   | { case: 'zarr-v3-no-multiscales' }
   | { case: 'zarr-query-error'; errorMessage?: string }
+  // Zarr - metadata is valid, but the layout will make viewing awkward
+  | { case: 'zarr-single-level'; size: string }
+  | { case: 'zarr-large-chunks'; size: string; compressed: boolean }
+  | { case: 'zarr-axis-order'; axisOrder: string; expectedOrder: string }
   // N5 - query never fired
   | { case: 'n5-has-s0-no-attrs' }
   | { case: 'n5-has-attrs-no-s0' }
@@ -71,6 +75,24 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
         description: variant.errorMessage
           ? `Could not read Zarr metadata. ${variant.errorMessage}`
           : 'Could not read Zarr metadata.'
+      };
+    case 'zarr-single-level':
+      return {
+        kind: 'warning',
+        title: 'Only one resolution level',
+        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Viewers must read full-resolution at every zoom level, so viewing the whole image is far more expensive than it needs to be. Generating a multiscale pyramid fixes this.`
+      };
+    case 'zarr-axis-order':
+      return {
+        kind: 'warning',
+        title: 'Axes are not in the order OME-Zarr specifies',
+        description: `The axes are ordered ${variant.axisOrder}, but the spec requires T, C, Z, Y, X order - here that would be ${variant.expectedOrder}. Many tools take the last two axes to be the image plane, so they will show a cross-section rather than the expected view. Rewriting the dataset with the axes in spec order avoids this.`
+      };
+    case 'zarr-large-chunks':
+      return {
+        kind: 'warning',
+        title: 'Chunks may be too large for efficient viewing',
+        description: `This dataset uses ${variant.size} chunks ${variant.compressed ? '(before compression)' : '(without compression)'}. Chunk files larger than the browser's cache limit are re-downloaded on every access, which makes viewing slow. A final chunk size of 1-10 MB works best.`
       };
     case 'n5-has-s0-no-attrs':
       logger.info(
--- a/frontend/src/omezarr-helper.ts
+++ b/frontend/src/omezarr-helper.ts
@@ -30,28 +30,41 @@ export type Metadata = OmeZarrMetadata & {
  */
 export type DatasetWarning =
   | { case: 'zarr-single-level'; size: string }
-  | { case: 'zarr-large-chunks'; size: string; compressed: boolean }
-  | { case: 'zarr-axis-order'; axisOrder: string; expectedOrder: string };
+  | {
+      case: 'zarr-large-chunks';
+      size: string;
+      compressed: boolean;
+      sharded: boolean;
+    };
 
 /**
- * Chunks above this defeat the browser cache. Applies when the array is stored
+ * Chunks above this are large enough to slow viewing down: a viewer has to
+ * fetch a whole chunk to show any part of it. Applies when the array is stored
  * without compression, so the size we compute is the size that transfers.
+ *
+ * Published guidance puts the useful range well below this - the image.sc
+ * discussion of OME-Zarr chunk sizes lands on 1-10 MB [1], AWS gives 8-16 MB as
+ * typical for S3 byte-range reads [2], and webknossos recommends 32^3 to 128^3
+ * voxel inner chunks [3]. 32 MB is the top of what anyone recommends, so a chunk
+ * past it is outside the range rather than merely on the large side of it.
+ *
+ * [1] https://forum.image.sc/t/should-compression-play-a-role-in-selecting-chunk-sizes-for-ome-zarr-v0-4-datasets/117877
+ * [2] https://d1.awsstatic.com/whitepapers/AmazonS3BestPractices.pdf
+ * [3] https://docs.webknossos.org/webknossos/data/zarr.html
  */
-export const MAX_CHUNK_BYTES = 10 * 1024 ** 2;
+export const MAX_CHUNK_BYTES = 32 * 1024 ** 2;
 /**
  * The same limit for compressed arrays, where all we can compute is the logical
  * (uncompressed) extent and the real transfer is some unknowable fraction of it.
- * Set high enough that a typical ratio cannot push a healthy dataset over.
+ * Doubled, so a chunk has to be outside the recommended range even at a
+ * conservative 2x ratio before we say anything.
  */
-export const MAX_LOGICAL_CHUNK_BYTES = 32 * 1024 ** 2;
+export const MAX_LOGICAL_CHUNK_BYTES = 64 * 1024 ** 2;
 /**
  * Below this, a single-level image is small enough that the missing pyramid
  * costs nothing worth mentioning.
  */
 export const MAX_SINGLE_LEVEL_BYTES = 1024 ** 3;
-
-/** The axis order OME-Zarr requires, minus any axes the dataset omits. */
-const CANONICAL_AXIS_ORDER = ['t', 'c', 'z', 'y', 'x'];
 
 /**
  * Whether the array's chunks are compressed on disk.
@@ -71,6 +84,18 @@ function hasCompressionCodec(codecs: NonNullable<Metadata['codecs']>): boolean {
     }
     return classifyCodec(codec.name) !== 'structural';
   });
+}
+
+/**
+ * Whether the array is sharded. Worth knowing because zarrita resolves the
+ * sharding codec when it opens the array - `arr.chunks` is then the inner chunk
+ * shape, the unit a viewer actually fetches, and not the shard around it. The
+ * warning says so, since "chunk" alone is ambiguous once sharding is in play.
+ */
+function isSharded(metadata: Metadata): boolean {
+  return (
+    metadata.codecs?.some(codec => codec.name === 'sharding_indexed') ?? false
+  );
 }
 
 function isStoredCompressed(metadata: Metadata): boolean {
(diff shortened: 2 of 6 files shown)


mkitti:
At some point, we may want to provide virtualization for Zarr datasets like this as well.

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
```diff
@@ -72,6 +76,24 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
           ? `Could not read Zarr metadata. ${variant.errorMessage}`
           : 'Could not read Zarr metadata.'
       };
+    case 'zarr-single-level':
```
mkitti: This single level issue should only be raised if the array is also large. A single level might be appropriate for a small 2D image.

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
```diff
@@ -72,6 +76,24 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
...
+      return {
+        kind: 'warning',
+        title: 'Only one resolution level',
+        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Viewers must read full-resolution at every zoom level, so viewing the whole image is far more expensive than it needs to be. Generating a multiscale pyramid fixes this.`
+      };
+    case 'zarr-axis-order':
+      return {
+        kind: 'warning',
+        title: 'Axes are not in the order OME-Zarr specifies',
+        description: `The axes are ordered ${variant.axisOrder}, but the spec requires T, C, Z, Y, X order - here that would be ${variant.expectedOrder}. Many tools take the last two axes to be the image plane, so they will show a cross-section rather than the expected view. Rewriting the dataset with the axes in spec order avoids this.`
```
mkitti: RFC 3 will likely make this advice obsolete with respect to the spec statement:

https://ngff.openmicroscopy.org/rfc/3/

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
```diff
@@ -72,6 +76,24 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
...
+      };
+    case 'zarr-large-chunks':
+      return {
+        kind: 'warning',
+        title: 'Chunks may be too large for efficient viewing',
+        description: `This dataset uses ${variant.size} chunks ${variant.compressed ? '(before compression)' : '(without compression)'}. Chunk files larger than the browser's cache limit are re-downloaded on every access, which makes viewing slow. A final chunk size of 1-10 MB works best.`
```
mkitti: We may need to qualify this for shards. Shards are also large top-level chunks.

frontend/src/omezarr-helper.ts
```diff
@@ -22,6 +24,173 @@ export type Metadata = OmeZarrMetadata & {
...
+  | { case: 'zarr-single-level'; size: string }
+  | { case: 'zarr-large-chunks'; size: string; compressed: boolean }
+  | { case: 'zarr-axis-order'; axisOrder: string; expectedOrder: string };
+
+/**
+ * Chunks above this defeat the browser cache. Applies when the array is stored
+ * without compression, so the size we compute is the size that transfers.
+ */
+export const MAX_CHUNK_BYTES = 10 * 1024 ** 2;
+/**
```
mkitti, line 56: I'm not sure if this limit is the one that would apply to Zarr chunks.

https://claude.ai/share/2c1c2b4f-c9a3-43d3-a24d-9de809b04062

mkitti (changes requested):
I am not sure how well these warnings are grounded in the spec and perhaps more specifically in Neuroglancer's implementation. I'm not sure where the 10 MiB cache limit comes from. Could we check the cache limits and add citations when possible to clearly state whether these are browser or middleware issues.

I suspect that the dimension order could be relaxed here especially in light of RFC-3. The main relevant ordering is probably the clustering of the spatial dimensions at the end but the order of the spatial dimensions is not really a problem for Neuroglancer.

Another warning I would add here is the dimensionality of the chunks versus the dimensionality of the array. If we have 2D chunks and a 3D spatial array, we will have issues for viewing in some of the orthogonal views. If the array is nD, then we want the dimension of the chunks to have non-singleton dimensions for at least`min(n, 3)` dimensions.

Also chunks that are too small might be an issue as well. In the extreme a chunk that is a single voxel is highly problematic. I would suggest a minimum element count of 256 elements if there are more elements than this in the array.

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
krokicki: I know that, but in the meantime it is creating issues, specifically with ome-zarr.js (and I would bet there are other cases). 

For example: https://github.com/BioNGFF/ome-zarr.js/blob/main/src/image.ts#L336-L337

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
krokicki: I'm not sure I follow. Shards are on the server side, the client side deals with chunks. Why would shard size matter to the client?

frontend/src/omezarr-helper.ts
krokicki: Some browsers do have a limit on the size of a file they will cache. For example, Firefox will only cache up to 5 MB files in memory (`browser.cache.memory.max_entry_size`) and 50 MB on disk (`browser.cache.disk.max_entry_size`). This 10 MB threshold is uncompressed, and assumes 2-3x compression will put the chunk under the 5 MB threshold. It's just a heuristic.

Of course, there are other reasons to favor smaller chunk sizes, but not too small. Maybe there are better heuristics or rules we can use here.

krokicki:
> I am not sure how well these warnings are grounded in the spec and perhaps more specifically in Neuroglancer's implementation. I'm not sure where the 10 MiB cache limit comes from. Could we check the cache limits and add citations when possible to clearly state whether these are browser or middleware issues.

Browsers have limits at 5 MB, as explained in the code comment. [This thread](https://forum.image.sc/t/should-compression-play-a-role-in-selecting-chunk-sizes-for-ome-zarr-v0-4-datasets/117877) suggests that reasonable chunk file sizes are between 1-10MB. 

What we want to do here is to provide a soft warning to people when they use chunk sizes that are clearly not ideal. Where would you set that threshold without creating too many false positives? I don't expect that this threshold can be a perfect classifier for correct/wrong, since there are many different (and often opposing) forces acting upon it.  

> I suspect that the dimension order could be relaxed here especially in light of RFC-3. The main relevant ordering is probably the clustering of the spatial dimensions at the end but the order of the spatial dimensions is not really a problem for Neuroglancer.

But RFC-3 is not a part of any official specification version, and I'm not sure when it would be merged. Even after it is released, it will take time for libraries (such as ome-zarr.js) to catch up with it. In the meantime, it creates issues when the spec is not followed. Although there may be good reasons for deviating from the spec (and maybe some of those inspired RFC-3), I don't see the problem with alerting people that they are deviating. 

> Another warning I would add here is the dimensionality of the chunks versus the dimensionality of the array. If we have 2D chunks and a 3D spatial array, we will have issues for viewing in some of the orthogonal views. If the array is nD, then we want the dimension of the chunks to have non-singleton dimensions for at least`min(n, 3)` dimensions.

Yes, but that seems out of scope for this PR, which is trying to address concrete issues that happened in production.

> Also chunks that are too small might be an issue as well. In the extreme a chunk that is a single voxel is highly problematic. I would suggest a minimum element count of 256 elements if there are more elements than this in the array.

Yes, but also out of scope here.

allison-truhlar:
Some minor suggestions for the wording of the warnings:
- Only one resolution level - "...for XX GB data. **Without multiple levels, viewers read the full-resolution data at every zoom level, making viewing slow.** Generating...
- Chunks may be too large for efficient viewing - "...(before compression). **This may exceed your browser's cache limit and cause slow viewing due to frequent chunk file re-downloads.** A final chunk size...

mkitti:
> Where would you set that threshold without creating too many false positives?

Amazon's "Best Practices Design
Patterns: Optimizing
Amazon S3 Performance" states

> Typical sizes for byte-range requests are 8 MB or 16 MB.

https://d1.awsstatic.com/whitepapers/AmazonS3BestPractices.pdf

32 MiB is the upper limit I usually recommend, especially in the absence of sharding. I typically recommend 256³ for uint16 datasets on microscope systems running Windows when I anticipate relatively high file overhead from the Windows file system and the pending remote transfer of the entire dataset.

1 MB chunk sizes is the target I recommend when on systems where the number of files is not an acute constraint.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-files-best-practices.html

> Using larger IO sizes (1 MB or more) amortizes per-operation overhead and is more cost effective than many small reads or writes.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-files-best-practices.html

My main recommendation for Zarr v3 is 1 GiB shards (hard upper limit is 2 GiB to support 32-bit systems that address memory or arrays via signed 32-bit integers, such as Java and JavaScript). Scalable Minds recommends 32³ to 128³ voxel chunks within the shard, preferring smaller chunks for smaller datasets.

https://docs.webknossos.org/webknossos/data/zarr.html

The Keller Lab, Fernando Amat, specifically, found optimal chunk sizes around 96 x 96 x 8 in shape for KLB.

https://www.nature.com/articles/nprot.2015.111/figures/9

So depending on the underlying systems and intended uses we recommendations spanning 32³ to 256³ voxels and 32 KiB to 32 MiB in terms of bytes.

It would be worthwhile to do some study on this with systematic benchmarks to make sure we understand the impact of chunk size on our current systems and the attribution of this across various layers and applications.

mkitti:
> Although there may be good reasons for deviating from the spec (and maybe some of those inspired RFC-3), I don't see the problem with alerting people that they are deviating.

Conversion and data copies can get expensive. If our main target application is Neuroglancer or other viewers intended for volumetric display (BDV, BigStitcher, Imaris), I do not see the cost of conversion or copies as being worthwhile here for permutations of the spatial dimensions within the lowest strided dimensions.

The original OME recommendations are based on viewing applications such as OMERO which focus on 2D display and are carried over from OME-TIFF.

My main concern here is that the chunks are 3D and isotropic. In that case, the permutation of XYZ should not strongly matter.

It seems that the main concern is for thumbnail generation. However, I find the assumption of which two dimensions to preview to be conceptually flawed. If we have a true isotropic volume, then a preview should be either 3D or allow for multiple 2D slices / projections (XY, YZ, ,XZ).

mkitti:
My question here is whether these issues are actually intrinsic to the array. In many cases, I think we could adapt the display to the array properties.

Do we only have one scale level or large chunks? Zoom-in to display fewer chunks at a time.

Are the chunks not isotropic? Orient the initial display for the optimal two dimensions. Perhaps do not show all three orthogonal dimensions if this would result in loading multiple chunks. If we have XY tiles with no Z depth, then only show the XY plane perspective.

krokicki:
> So depending on the underlying systems and intended uses we recommendations spanning 32³ to 256³ voxels and 32 KiB to 32 MiB in terms of bytes.

I will increase the chunk size threshold for warnings from 10 MB to 64 MB, which should account for compression down to ~32 MB. That would still catch the 128 MB chunk issue we saw, while accommodating the larger chunks you use in practice. I'd also be interested in understanding the cases where those larger chunk sizes are beneficial, since that could help inform the guidance we give.

> I find the assumption of which two dimensions to preview to be conceptually flawed.

I agree that there are limitations in this implementation, but it does follow the existing specification, so for now I think it makes sense to keep it as-is. When the spec is changed (e.g. when RFC-3 is merged) then we will need to revisit how thumbnails are implemented. 

> My question here is whether these issues are actually intrinsic to the array. In many cases, I think we could adapt the display to the array properties.

For this PR I'm trying to keep the scope focused on providing practical advice that can be useful to users today. It would be helpful to capture some of the adaptive-display ideas separately, so that we can consider them independent of this PR.

mkitti:
Here's the OME-Zarr to roadmap to v1.0:
https://forum.image.sc/t/our-proposed-roadmap-to-ome-zarr-1-0/121995/1

image

RFC-3 related comments:
1. "RFC-3 (Relaxing dimension restrictions) almost ready for review"
2. "For example, using soon-to-be-released versions of napari-ome-zarr and napari, you can now read [example RFC-3 data](https://github.com/clbarnes/ome-zarr-rfc3-data/pull/1#issuecomment-4963532700) into napari and get accurate axis metadata, and [work is ongoing on bioformats2raw](https://github.com/glencoesoftware/bioformats2raw/pull/330) to accurately export additional axes to RFC-3 compliant OME-Zarr."

As shown in the roadmap, this is the focus of primary activity in the community right now. The objective is to finalize RFC-3 this month.

> For this PR I'm trying to keep the scope focused on providing practical advice that can be useful to users today.

This is what is confusing me. I do not see the axis order advice as practical. It does not affect Neuroglancer. I do not think people should waste time or resources on axis order issues.

Even worse, I think the axis order advice is suboptimal for many applications. For example, if you have a true RGB image or RGBA image often these are expected to be the last dimension as exemplified by these libraries:

1. Matplotlib: [`matplotlib.pylot.imshow`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html) takes arrays in `(M, N, 3)` or `(M, N, 4)` order.
2. Tensorflow: [`tf.nn.conv2d`](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d) - With the default format "NHWC", the data is stored in the order of: batch_shape + [height, width, channels].
3. scikit.image [`skimage.color`](https://scikit-image.org/docs/stable/api/skimage.color.html)
4. [ImageJ](https://imagej.net/ij/docs/guide/146-26.html) "Interleaved RGB images have pixels stored contiguously (rgbrgbrgb...) in a single image plane. Planar RGB images have the red, green and blue image data stored in separate 8-bit sample planes. ImageJ saves RGB images (both TIFF and raw) in interleaved format."

While it is true that many images coming our way are from fluoresence microscopy are not in RGB format, fluorescence microscopy is the exception. Most histology images (e.g. H&E sections) for example are in RGB format.

Regarding ome-zarr.js, I do see a lot of code that does not assume the order of the dimensions:
1. https://github.com/BioNGFF/ome-zarr.js/blob/37c4ed5037902d0a4999bf207debf53710e4f4e6/src/render.ts#L153
2. https://github.com/BioNGFF/ome-zarr.js/blob/37c4ed5037902d0a4999bf207debf53710e4f4e6/src/image.ts#L237

I submitted a pull request to address the parts of the code where the order of the dimensions is assumed:
https://github.com/BioNGFF/ome-zarr.js/pull/47

I would particularly focus on the chunk issues for the warnings since there really is not any solution that we can implement short of rewriting the entire array. The single resolution issue can be resolved by adding the additional scales by materializing them to disk or calculating them on the fly in some cases, which is a service we might be able to provide.

The dimension order issue is at best superfluous and at worse deleterious to potential applications. The dimension order issue should just be left to a OME-NGFF validator and hopefully removed from the spec as soon as possible. If a tool assumes the last two dimensions are Y and X despite clear metadata to contrary, it is a bug that we can and should fix.

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
mkitti: A shard is also a chunk. It is just a specific kind of chunk that contains other "inner" chunks. To specify the size of a "shard" in Zarr I say `chunk_shape: [1024, 1024, 1024]`.

Your warning shoud be qualified to say "inner" chunk perhaps.

frontend/src/omezarr-helper.ts
mkitti: I would suggest either softening the warning copy to something like "may exceed some browsers' in-memory HTTP cache limits, causing slower (disk-cache or re-fetch) access" rather than stating a re-fetch as the consequence, or dropping the specific causal mechanism and just framing the threshold as a general chunk-size-for-viewing-performance heuristic (which is closer to where the 64 MB / 32 MB revision in this thread is already headed, and is more defensible than pinning it to one browser's memory-tier default).

krokicki:
@mkitti Thanks for making that PR to ome-zarr.js. Although I still think that OME-Zarr users should respect the spec as written, I've dropped the axis-order warning since we can't find consensus about it. I updated the thresholds to the numbers you recommended. I also added your "inner chunks" terminology when data are sharded.

@allison-truhlar Thanks, I adapted your wording with Mark's feedback.

frontend/src/components/ui/BrowsePage/MetadataHint.tsx
```diff
@@ -72,6 +80,18 @@ function getHintConfig(variant: MetadataHintVariant): HintConfig {
           ? `Could not read Zarr metadata. ${variant.errorMessage}`
           : 'Could not read Zarr metadata.'
       };
+    case 'zarr-single-level':
+      return {
+        kind: 'warning',
+        title: 'Only one resolution level',
+        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Without multiple levels, viewers read the full-resolution data at every zoom level, making viewing slow. Generating a multiscale pyramid fixes this.`
```
mkitti, line 87:
```suggestion
        description: `This dataset declares multiscales but only supplies a single level, for ${variant.size} of data. Without multiple levels, viewers read the full-resolution data at every zoom level, making viewing slow for a large field of view. Generating a multiscale pyramid fixes this.`
```

mkitti: Viewing the dataset when zoomed in to a small field of view is unaffected by the lack of a pyramid.

I usually create such datasets when initially copying data from an instrument computer to shared storage. The copying and pyramid generation as distinct steps.

mkitti (approved):

PR #431, opened 2026-08-19
feat: Neuroglancer Views polish (PR 7, ngviews-07-cart-and-viewer-polish)

allison-truhlar:
Follow-up polish on the Neuroglancer Views stack, addressing three issues raised in the last dev meeting round.

## What changed

- **Cart rows expand without a Data Link.** `CartDatasetRow` now fetches OME-Zarr metadata and channel labels from the internal `/api/content/<fsp>/<path>` URL (credentialed), so every valid Zarr dataset in the cart can be inspected before a View is created. Previously the "Channels load after the View is created" hint appeared for any dataset that didn't already have a proxied path, which was inconsistent — the same dataset with an existing Data Link had full expansion. The proxied-path lookup and its `useAllProxiedPathsQuery` dependency drop out of `CartList` entirely.
- **Embedded viewer header collapses to one row.** The redundant `h6` title (already shown in the breadcrumb) is gone, and the action buttons — Copy link, Download JSON, Open external, Fullscreen — move up to sit on the breadcrumb row. Gives the iframe more vertical space.
- **NG Views table: resizable Sources column, horizontal scroll, one row per entry.**
  - Fixes the missing 6th grid track so the Actions cell stops wrapping to a second row on each entry.
  - Adds a `gridColsStyle` prop on `TableCard` (inline `gridTemplateColumns`) as an alt to the static `gridColsClass`, since Tailwind JIT can't emit arbitrary-width classes at runtime. Existing tables keep their `gridColsClass` and are unaffected.
  - The Sources column has a drag handle at the column's right edge that mirrors the file browser's resize UI (thin line, primary color on hover/drag).
  - Each source path renders on its own truncated line with a native `title` tooltip on hover — links remain clickable and navigate to the file browser.
  - The header + body are wrapped in `overflow-x-auto` so widening Sources doesn't push Sharing, Updated, and Actions off-screen out of reach.
  - Every cell gets explicit `justify-start` + `text-left`.

## Verified

- `pixi run node-check` — no new type errors (pre-existing failures in unrelated files: `ContextMenu.tsx`, `PathFormatOptions.tsx`, `PermissionsTable.tsx`, `sshKeyQueries.ts`, `ColorsPageSync.test.ts`).
- `pixi run node-eslint-check` — 0 errors, only a pre-existing unused-import warning in `NeuroglancerOptions.tsx`.
- `pixi run test-frontend` — 384 tests pass. `CartList.test.tsx`, `CartTab.test.tsx`, and `ngViewsColumns.test.tsx` updated for the new URL source and the resizable-column hook signature.

## Not verified

Manual dev-server browser check. The changes are visual, so a quick eyeball pass on the three affected surfaces is worthwhile before merge.


krokicki (changes requested):
This looks great. There are just a few issues I think we should address before merging:
1) There is currently an error when you attempt to delete a data link that is part of a view. This only happens on the dev server (maybe it's Postgres-specific?)
2) On my own server (where the above bug is not present) I can delete a data link, but the View row in the Views table continues to look normal. It would be nice to mark it broken somehow, so the user knows what to expect before opening it.
3) I think it would be useful for the View to track what FSP/subpath it is linked to, in addition to the data link. That will let us repair broken views in the future, if we decide to do that (which needs further discussion), or allow users to navigate directly to the folder from the View.
4) The right-click "View in Neuroglancer" and "Add to Neuroglancer cart" actions are confusing to me. It seems that they appear on every folder, so you can create broken views that don't work from the outset. I don't think it's possible to follow the usual logic here because we would need to scan all children for `.zattrs`/`zarr.json`. One idea is to change the action to just a generic "add to cart", and then in the cart the user can decide what to do with the items. For the same reason, I would remove the "New view from selection" button in the bottom overlay.

allison-truhlar:
> This looks great. There are just a few issues I think we should address before merging:
> 
>     1. There is currently an error when you attempt to delete a data link that is part of a view. This only happens on the dev server (maybe it's Postgres-specific?)
Fixed in commit https://github.com/JaneliaSciComp/fileglancer/pull/454/commits/ac6efd0609414bc1f2b764e8056ba6fb4276e5e2, in PR #454 

>     2. On my own server (where the above bug is not present) I can delete a data link, but the View row in the Views table continues to look normal. It would be nice to mark it broken somehow, so the user knows what to expect before opening it.
Commits https://github.com/JaneliaSciComp/fileglancer/pull/454/commits/290a01c9b202d6711a0dac4497ccca1cafc6188c and https://github.com/JaneliaSciComp/fileglancer/pull/454/commits/290a01c9b202d6711a0dac4497ccca1cafc6188c keep broken sources in the Views table and adds a broken link icon next to the broken source. A tooltip when hovering on the icon explains that the data link for the source has been deleted, so that source will no longer appear in the View.

>     3. I think it would be useful for the View to track what FSP/subpath it is linked to, in addition to the data link. That will let us repair broken views in the future, if we decide to do that (which needs further discussion), or allow users to navigate directly to the folder from the View.
Done in commits https://github.com/JaneliaSciComp/fileglancer/pull/454/commits/cee07b1ce9224897408b788cb8f49bb5a9dd1c1e and https://github.com/JaneliaSciComp/fileglancer/pull/454/commits/6ecffed404aabc980582629daca292e8a1e8f6a7. Added to the UI in commit https://github.com/JaneliaSciComp/fileglancer/pull/455/commits/2493ca4582a525633ee4e5912db9263b31d8bb35.

>     4. The right-click "View in Neuroglancer" and "Add to Neuroglancer cart" actions are confusing to me. It seems that they appear on every folder, so you can create broken views that don't work from the outset. I don't think it's possible to follow the usual logic here because we would need to scan all children for `.zattrs`/`zarr.json`. One idea is to change the action to just a generic "add to cart", and then in the cart the user can decide what to do with the items. For the same reason, I would remove the "New view from selection" button in the bottom overlay.
Commit https://github.com/JaneliaSciComp/fileglancer/pull/455/commits/a9ec89257ff75e5421dcea80a4482dc67b71368c drops "View in Neuroglancer" from row menu and renames "Add to Neuroglancer cart" to just "Add to cart". Commit https://github.com/JaneliaSciComp/fileglancer/pull/455/commits/26a7312e9bb8c83ef98272a3fcfd9e7b9305be7b makes the cart the only path to a view, dropping the "New View" button from the floating bottom menu bar.

PR #442, opened 2026-09-02
Fix client validation and API Tokens page link

krokicki:
## Summary
- Strip whitespace from `FILEGLANCER_URL`/`FILEGLANCER_TOKEN` (and constructor args) before use, and require an `http://`/`https://` scheme, so bad input fails fast with a clear message instead of an opaque `httpx`/`h11` error on the first request.
- Fix the "authoring guide" link on the API Tokens page, which pointed at the PyPI package page and rendered as "authoring guidePython package." — now reads "Python package" and links correctly.

Follow-up to #435, whose branch was deleted after merge.

## Test plan
- [x] Manually verified `Fileglancer(url=..., token=...)` raises `FileglancerError` for a scheme-less URL, a typo'd scheme, and succeeds with whitespace-padded url/token.
- [x] Frontend Prettier/ESLint checks pass (via pre-push hook).

### DIFF
--- a/fileglancer/client.py
+++ b/fileglancer/client.py
@@ -44,11 +44,15 @@ class Fileglancer:
 
     def __init__(self, url: Optional[str] = None, token: Optional[str] = None,
                  timeout: float = 60.0):
-        url = url or os.environ.get("FILEGLANCER_URL")
-        token = token or os.environ.get("FILEGLANCER_TOKEN")
+        url = (url or os.environ.get("FILEGLANCER_URL") or "").strip()
+        token = (token or os.environ.get("FILEGLANCER_TOKEN") or "").strip()
         if not url:
             raise FileglancerError(
                 "No Fileglancer server URL. Pass url= or set FILEGLANCER_URL.")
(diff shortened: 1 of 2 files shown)


fileglancer/client.py
```diff
@@ -44,11 +44,15 @@ class Fileglancer:
...
-        token = token or os.environ.get("FILEGLANCER_TOKEN")
+        url = (url or os.environ.get("FILEGLANCER_URL") or "").strip()
+        token = (token or os.environ.get("FILEGLANCER_TOKEN") or "").strip()
         if not url:
             raise FileglancerError(
                 "No Fileglancer server URL. Pass url= or set FILEGLANCER_URL.")
+        if not url.startswith(("http://", "https://")):
+            raise FileglancerError(
+                f"Invalid Fileglancer server URL {url!r}: must start with "
+                "http:// or https://.")
```
mkitti, line 55: This assumes that `url` and `token` are string, but no conditional statement has clearly established that before we call `strip()`.

PR #443, opened 2026-09-02
Dial an app service with the scheme it published

mkitti:
Dials an app service with the scheme it published, so a service that terminates TLS itself is reached over HTTPS instead of being sent a plaintext request. Stacked on #440.

**Deployment half: JaneliaSciComp/fileglancer-hub#16** (four lines of nginx). Both must land; each is inert without the other.

## Why this is a prerequisite, not a hardening step

#440 discards the scheme: `upstream_from_service_url` keeps only `host:port`, and the vhost hardcodes `proxy_pass http://$upstream`. That looks like it only leaves the last hop cleartext. It is worse than that — two apps already serve HTTPS behind Caddy and already publish an `https://` URL to `$SERVICE_URL_PATH`:

- **`JaneliaSciComp/marimo_ai_sandbox`** — `container/caddy-lib.sh` generates an `openssl` self-signed leaf (SANs: node FQDN, short hostname, `localhost`, host IPs); `container/https-wrap.sh:262` publishes `https://<host>:<port>/?access_token=…`. It bypasses `auto_url` deliberately so Fileglancer does not publish the backend's plain-HTTP port instead of Caddy's TLS one.
- **Another shipped app** — Caddy `tls internal` on a fixed port 8443, publishes `https://$(hostname -f):8443/`.

Sending a plaintext request to a TLS listener does not quietly lose encryption; the listener refuses it. So setting `apps.service_proxy_domain` today breaks both apps outright. Reproduced in the harness below: with the scheme hardcoded, the TLS upstream answers `400 The plain HTTP request was sent to HTTPS port`.

## What's in it

`GET /api/apps/resolve` now also answers `X-Fg-Upstream-Scheme: http|https`, and the vhost does `proxy_pass $upscheme://$upstream`.

**A second header rather than a prefix on the existing one.** Folding the scheme into `X-Fg-Upstream` would be one fewer line of nginx, but it makes an old proxy config build `proxy_pass http://https://host:port`. With a separate header, a proxy configured before this existed ignores what it does not know and keeps dialing `http` — exactly today's behavior — which matters given #440 and hub#15 already have a release-ordering constraint between them.

**`upstream_from_service_url` is untouched.** It is the header-injection and `proxy_pass`-interpolation gate and its shape is worth not disturbing, so the scheme comes from a sibling function that the endpoint calls only *after* the authority has been accepted. Anything not exactly `https` yields `http`: guessing `https` at a plaintext listener turns a working service into a failed handshake, whereas the reverse is the status quo. Nothing needs refusing either — `read_service_url_file` (`fileglancer/apps/jobfiles.py:164`) already rejects a URL that is not `http://` or `https://` before it can reach the database.

**The scheme is cached with the upstream.** This is the subtle one. Hits outnumber misses by design, so deriving the scheme on the miss path alone would leave an HTTPS service dialed as plaintext for the rest of the TTL, flapping between the two as entries expire. There is a regression test for exactly that.

**Observability instead of a policy knob.** A `plaintext` label is counted *beside* `hit` and `miss` rather than instead of them, so the once-a-minute totals read as a fraction: `hit=3980 miss=17 plaintext=233 cached=9`. An `apps.service_proxy_upstream_tls: allow | require` setting was floated on #440 and withdrawn — `require` would 403 every plaintext app, and TensorBoard has no TLS option at all, which left the setting with one legal value. **No new settings.**

Frontend: no changes. `build_proxied_service_url` already forces `https` for the browser URL, so the published link is unaffected.

## What this deliberately does not do

**It does not verify the certificate.** `proxy_ssl_verify` is `off`. Those certificates are generated on the compute node at launch and signed by nobody: there is no trust anchor to check them against, and the node and port change every launch, so there is nothing stable to pin. This is opportunistic encryption — it keeps the service token off the wire in cleartext and authenticates nothing, which is no worse than the plaintext hop it replaces and no better. `docs/ServiceProxy.md` now says that in those words rather than implying verified TLS.

The options for actually authenticating the upstream — self-signed with pinning, an internal CA, a uniform TLS sidecar — are recorded in `docs/superpowers/specs/2026-09-02-app-service-upstream-tls-options.md` along with why each is more than a mostly trusted intranet warrants today. Worth reading before anyone proposes one; two are dead ends for specific reasons (`proxy_ssl_trusted_certificate` accepts no variables, and nginx has no upstream-fingerprint directive).

**It does not change `auto_url`.** `_build_service_url_publisher` still emits `http://`. Both motivating apps write their own URL, and no app both uses `auto_url` and terminates TLS. If one ever does, the smallest change is a manifest boolean feeding that `printf`.

## Testing

`pixi run -e test test-backend`: **930 passed, 77 skipped.** New coverage: the scheme reported for both `http` and `https` upstreams, the fail-safe fallback for anything unparseable, the `plaintext` tally landing beside the hit or miss, and the cache-downgrade regression.

**nginx, verified live rather than assumed.** `proxy_pass` taking a scheme from a variable was the one load-bearing assumption, so I stood up a harness where a single nginx 1.24 instance plays every role — a plaintext upstream, a self-signed TLS upstream, a stand-in resolve endpoint answering the two headers, and the real vhost:

```
job-1 (published http):  code=200 body=PLAINTEXT-UPSTREAM
job-2 (published https): code=200 body=TLS-UPSTREAM
job-9 (no upstream):     code=403
```

With the scheme hardcoded back to `http://`, the same harness gives `job-2 (https app): code=400`, `plain HTTP request was sent to HTTPS port`. Both the bug and the fix are demonstrated rather than argued. The real hub `nginx.conf` with the change also passes `nginx -t` in a wrapper config, matching hub#15's own standard.

Not verifiable here: whether each app behaves end to end behind the proxy still needs the DNS record and certificate from hub#15's checklist.

## Known limitations, recorded rather than hidden

- **The other Caddy-fronted app will still 404 through the proxy**, scheme or no scheme. Its Caddyfile site address is Host-matched (`{$CADDY_HOSTNAME}:8443, localhost:8443, 127.0.0.1:8443`) and the vhost passes `Host $host` through unchanged — load-bearing for JupyterLab's WebSocket origin check — so Caddy sees `job-<id>.services…`, matches no site, and refuses. That is a one-line fix in that repo (a catch-all `:8443`), and its GitHub OAuth callback is pinned to a fixed host *and port* besides, so it may want a way to opt out of being republished entirely. There is no such manifest option today. `marimo_ai_sandbox` uses a port-only site address and is unaffected, so this PR claims that app works, not both.
- **The app-author-facing note is not here** because `docs/AuthoringApps.md` has migrated to `JaneliaSciComp/fileglancer-docs` (`src/content/docs/authoring/services.mdx`). Someone should add "if your service terminates TLS, publish an `https://` URL and the proxy will use it; a self-signed certificate is fine" there. That PR is not open.

@krokicki @JaneliaSciComp/fileglancer

### DIFF
--- a/docs/ServiceProxy.md
+++ b/docs/ServiceProxy.md
@@ -38,7 +38,7 @@ Leave `service_proxy_domain` empty to disable; the direct `http://<node>:<port>`
 
 ## Reverse proxy configuration
 
-Fileglancer does not proxy the traffic itself. It exposes `GET /api/apps/resolve`, which reads the `Host` header and answers `204` with `X-Fg-Upstream: <host>:<port>`, or `403`. The reverse proxy resolves each request through it and connects to the upstream directly, so no proxied bytes pass through the application server.
+Fileglancer does not proxy the traffic itself. It exposes `GET /api/apps/resolve`, which reads the `Host` header and answers `204` with `X-Fg-Upstream: <host>:<port>` and `X-Fg-Upstream-Scheme: http|https`, or `403`. The reverse proxy resolves each request through it and connects to the upstream directly, so no proxied bytes pass through the application server.
 
 Add a server block for the wildcard zone. This assumes a `map $http_upgrade $connection_upgrade` block already exists at the http level:
 
@@ -61,11 +61,18 @@ server {
   location / {
     auth_request     /_fg_resolve;
     auth_request_set $upstream $upstream_http_x_fg_upstream;
+    auth_request_set $upscheme $upstream_http_x_fg_upstream_scheme;
 
     # Required because proxy_pass targets a variable. Use whatever resolver the
     # host actually runs; 127.0.0.53 is systemd-resolved's stub.
     resolver 127.0.0.53 valid=30s;
-    proxy_pass http://$upstream;
+    proxy_pass $upscheme://$upstream;
+
+    # An app that terminates TLS does so with a certificate it generated on the
+    # compute node, so there is no trust anchor to verify against. See "TLS to
+    # the app host" below. Leave proxy_ssl_name at its default.
+    proxy_ssl_verify      off;
+    proxy_ssl_server_name on;
 
     proxy_set_header Host              $host;
     proxy_set_header X-Real-IP         $remote_addr;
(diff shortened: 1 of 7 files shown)


lefthook.yml
```diff
@@ -6,7 +6,10 @@ pre-push:
   jobs:
     - name: Run Prettier check; fix if check fails
       root: frontend/
-      run: pixi run ./prettier-check-and-fix.sh
+      # Invoked through bash explicitly: on Windows, pixi hands the script to
+      # the OS to execute and gets `os error 193` (not a valid Win32
+      # application), so the hook could never run there.
+      run: pixi run bash ./prettier-check-and-fix.sh
```
mkitti: We could drop this since we are dropping Windows support, but it is also not a bad idea to explicitly use bash.

mkitti:
This was determined to be not be necessary in the data environment.

PR #444, opened 2026-09-03
Sign service proxy hostnames so they can't be enumerated

krokicki:
## Summary
- Job ids are a small global sequence, so a bare `job-<id>` proxy hostname let anyone who could reach the proxy sweep `job-1`..`job-500` and find (and, for apps like TensorBoard that enforce no token of their own, reach) every running service on the instance.
- The hostname label now carries a MAC over the id, keyed by `session_secret_key`: `job-12-k7m2q9xr.services.example.org`. The MAC is verified before the database is touched at all, so resolution stays one indexed read with no new column.
- `session_secret_key` becomes required once `apps.service_proxy_domain` is set — an unset key is generated at random per process, which would break resolution across `uvicorn --workers N`, so the server now refuses to start instead of failing unpredictably at request time.

## Test plan
- [x] `pixi run -e test test-backend` (existing service-proxy tests updated for signed hostnames)
- [ ] Manually verify a service URL published through the proxy resolves with the new `job-<id>-<mac>` label, and that a tampered/guessed label gets a 403


krokicki:
This branch incorrectly merged in commits from another branch. Closing in favor of continuing this work in #440 .

PR #445, opened 2026-09-04
Let a service opt out of the HTTPS proxy

krokicki:
Adds `service_proxy: false` on a service entry point, for a service that cannot work when Fileglancer republishes it at a per-job HTTPS subdomain.

Docs half: JaneliaSciComp/fileglancer-docs#25. Independent — each is useful without the other.

## Why

The proxy rewrites every running service's URL to `https://job-<id>-<mac>.<zone>/...`. Most services do not notice. One registers its GitHub OAuth callback against a fixed host **and port**, so at a rewritten address the callback does not match and the app is unusable. There was no way for a manifest to say so, so enabling `apps.service_proxy_domain` broke that app with no recourse short of unregistering it.

This is the narrow fix for that, and only that. It is not a route to running services unencrypted by preference — the docs and the field description both say so explicitly.

## What's in it

`service_proxy: bool = True` on `AppEntryPoint`, snapshotted onto the job row at submit time, enforced in two places.

**Snapshotted rather than read from the manifest.** Same reasoning as `command` and `container` already: the decision must not change under a job that is already running, and the resolve endpoint must not have to parse a manifest to answer.

**Enforced at resolve as well as at publish.** Suppressing the published URL alone would make the opt-out advisory. The `job-<id>-<mac>` label is unguessable but derivable by anyone holding `session_secret_key`, so `/api/apps/resolve` refuses an opted-out job outright and counts it as `refused_proxy_disabled`, beside the existing refusal reasons rather than folded into them.

**Validated service-only, matching `auto_url`.** The check keys on the value being `false`, not on the field being set: `model_dump` writes the `True` default onto every entry point as manifests round-trip worker → server → DB cache, so a round-tripped `type: job` manifest would otherwise fail to revalidate. There is a test pinning exactly that.

**Migration.** One `add_column`, non-null with a `sa.true()` server default so existing rows backfill rather than leaving a nullable tri-state that every read site has to interpret. Verified on a scratch database: applies on top of `c3e9b7f41a28`, backfills a pre-existing row to true, and downgrades cleanly.

No new settings. No frontend changes — the rewrite already happens server-side in the job detail endpoint.

## Testing

`pixi run -e test pytest tests/`: **1026 passed.**

New coverage: the field defaults true, is rejected on a `type: job` entry point, survives a `model_dump` round trip, keeps the direct URL published on a server where the proxy is on, and is refused by `/api/apps/resolve` with the right counter label.

Migration checked by hand against a scratch SQLite database, both directions.

### DIFF
--- a/docs/ServiceProxy.md
+++ b/docs/ServiceProxy.md
@@ -36,6 +36,12 @@ Leave `service_proxy_domain` empty to disable; the direct `http://<node>:<port>`
 
 `session_secret_key` is required when the proxy domain is set, and the server refuses to start without it. An unset key is generated at random per process, so under `uvicorn --workers N` each worker would sign hostnames with a different key and most proxied requests would be refused. Rotating it invalidates live service URLs, on top of the session revocation rotation already causes.
 
+## Apps that opt out
+
+An individual service can decline to be republished by setting `service_proxy: false` on its entry point. Its direct `http://<node>:<port>` URL is published instead, and the resolve endpoint refuses its hostname, so it cannot be reached through the proxy even by someone who can derive the signed label. Refusals are counted as `refused_proxy_disabled` in the aggregate resolve log line.
+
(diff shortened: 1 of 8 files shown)


mkitti:
Please redact mention of mhs-spec-review which is not supposed to publicly disclosed and will likely drop fileglancer support.

https://github.com/JaneliaSciComp/marimo_ai_sandbox is public and will require this without #443 to handle its secrets heavy workload.

PR #446, opened 2026-09-05
fix: make the ~/.fileglancer state directory owner-only

krokicki:
## Problem

Fileglancer kept live credentials in three places where other people could read them.

**Job work directories and the app repo cache.** Everything under `~/.fileglancer` was created with the default umask, which on a shared filesystem means world-readable. A service job's work directory contains `service_url`, and that URL carries the access token the service authenticates with; services routinely echo the same URL into `stdout.log` as well. Any user who could read those files could connect to the service and act as its owner. The repo cache under `apps/` is a lesser case of the same thing: clones that may come from private repositories.

**The CLI's SQLite database.** `fileglancer start` created `~/.local/share/fileglancer` with the default umask too, and that database stores session ids verbatim — working login cookies — not just metadata.

**The `jobs.service_url` column.** It held the whole published URL, token and all, and nothing ever read that back. Its only consumer is `/api/apps/resolve`, which keeps the authority and discards the rest; the URL the browser gets from `GET /api/jobs/{id}` comes from that request's own worker read, with no fallback to the column. In a multi-user deployment the database is MySQL, which no file permission reaches, so database access, dumps and backups were a way to recover every running service's token.

## Fix

**The state directory.** `ensure_private_dir()` creates a directory 0700 and then locks every level up to and including the innermost `.fileglancer` above it.

The directory mode is what has to carry this. Chmodding individual files would mean chasing every writer — apps that write `SERVICE_URL_PATH` themselves when `auto_url` is off, and the scheduler writing the job's logs — and would still leave `stdout.log` exposed. One private directory covers all of them.

It is called from the four places the tree is created or first read, all of which already run as the owning user:

- `_action_submit` — the job work directory
- `_ensure_repo_cache` — the repo clone, above the cache-hit check rather than only on the clone path, so it is reached by every user rather than only by those cloning something new
- `_create_snapshot` — the pinned-commit checkouts
- `ensure_repo_snapshot`'s hot path — launching a pinned app reaches neither of the previous two, so for a user whose snapshots all predate this it is the only call that runs

Because it walks up to the state root, the first of those also locks down subtrees left world-readable by an earlier version. It self-heals when a user next launches an app or loads a manifest; nobody has to sweep anything, which matters because root cannot chmod these under root-squash NFS.

The `auto_url` publisher additionally writes `service_url` under `umask 077` and chmods it afterwards, so the file is 0600 on its own merits if it ever outlives its directory.

**The CLI database.** The directory is now 0700, which also covers the `-wal` and `-shm` sidecars SQLite writes beside the database, and the file is 0600 so a copy of it stays private. Both are unconditional, so an install created before this, or under a laxer umask, is repaired on the next start.

**The column.** `set_job_service_url` now stores `scheme://host:port`. Normalizing in the setter rather than at its one call site keeps the token out of the column whoever writes it. Migration `e4d1b7a92c58` rewrites existing rows so tokens already stored do not outlive the fix.

## Notes

- **This carries a data migration**, unlike the rest of the change. It rewrites `jobs.service_url` in place and nulls any row with no usable authority; a nulled row repopulates from the work directory on the next detail-page load. Verified with a full `alembic upgrade head` from an empty database, not just by calling the helper. Earlier revisions of this PR were aimed at a `3.1.1` backport — that is no longer the plan, and everything here ships from main in the next release.
- An authority carrying userinfo stores nothing rather than an origin. That is a credential too, and `upstream_from_service_url` refuses userinfo by design, so stripping it to make the row dialable would have quietly widened the SSRF gate.
- **`executor: local` is now documented as single-user only.** `_poll_local_jobs` runs in the server process and reads `job.pid` and `exit_code` straight from the work directory, so a 0700 work directory would leave another user's job in PENDING forever — the zombie timeout does not catch it, because the job has a `cluster_job_id`. That configuration is not supported, so the config template says so next to the option, and the upgrade path (move both reads into a worker action, the way `read_job_file` already does) is recorded next to the code that would have to change.
- A chmod that fails now explains itself. Failing to lock the leaf raises with a message naming what could not be made private; an ancestor we cannot repair (owned by another uid) is logged as a warning and ends the walk, rather than failing a job submission over a directory whose contents are already covered.
- `fileglancer debug` reading another user's job directory in-process now gets `None` instead of contents. It already swallows `OSError`, and root-squash NFS denied this in production anyway.
- `ensure_private_dir` locks only the leaf when handed a path outside a state directory — otherwise it would dictate where callers may create directories, and the snapshot tests relocate the cache base. The state root is found lexically rather than by resolving the path: a relocated state directory (a symlinked `~/.fileglancer`) resolves to somewhere with no `.fileglancer` component at all, so resolving would drop the walk entirely and leave the real tree unlocked.
- Already-issued tokens are not rotated by any of this, but no services are running, so there is nothing live to re-key.

## Testing

`pixi run -e test test-backend` — 1056 passed.

New coverage: `TestPrivateStateDir` (the whole chain is locked for a job dir, a repo clone and a snapshot dir; an already-755 tree is repaired; `$HOME` above the state root keeps its mode; the walk stops at the innermost state root; an unfixable ancestor warns while still locking the leaf; an unlockable leaf raises a readable error), the snapshot hot path repairing a world-readable tree, the publisher asserting 0600 both for a fresh file and one that already existed 0644, `service_url_origin` against the upstream gate including a header-injection attempt, and `tests/test_service_url_migration.py` for the rewrite, its idempotency, and that a stripped row still resolves to the same upstream.

Each of the regression tests was confirmed to fail with its fix reverted.

The database permissions are the one thing with no automated test — there is no CLI test harness, and standing one up to exercise three straight-line statements is more scaffolding than the fix. The resulting modes and SQLite's willingness to open the pre-created file were verified by hand.

### DIFF
--- a/fileglancer/apps/jobfiles.py
+++ b/fileglancer/apps/jobfiles.py
@@ -7,6 +7,8 @@ import shutil
 from pathlib import Path
 from typing import Optional
 
+from loguru import logger
+
 from fileglancer import database as db
 from fileglancer.settings import get_settings
 
@@ -49,18 +51,36 @@ def ensure_private_dir(path: Path) -> None:
     Every level from ``path`` up to and including ``~/.fileglancer`` is locked
     down, so the state root stays private however it was first created and
     subtrees made before this was enforced are covered too. Nothing above the
-    state root is touched.
+    state root is touched. An ancestor that cannot be chmod'ed (owned by another
+    uid) is warned about rather than raised, since the leaf is private either
+    way; failing to lock the leaf itself does raise.
 
     Must run as the owner of the tree (in the user worker, or as the user in CLI
     mode). Chmod rather than ``mkdir(mode=...)``: that mode is masked by the
     umask, and ignored entirely when the directory already exists.
     """
     path.mkdir(parents=True, exist_ok=True)
-    os.chmod(path, 0o700)
+    try:
+        os.chmod(path, 0o700)
+    except OSError as e:
+        raise PermissionError(
+            f"Cannot make {path} private, so it is not safe to write job "
+            f"credentials there: {e}"
+        ) from e
     for parent in path.parents:
         if _STATE_DIRNAME not in parent.parts:
             break  # above ~/.fileglancer (or outside it): not ours to touch
-        os.chmod(parent, 0o700)
+        try:
+            os.chmod(parent, 0o700)
+        except OSError as e:
+            # A state root owned by someone else is not ours to repair, and the
+            # directory we just locked is the one that holds the secrets. Say so
+            # and stop, rather than failing the caller over an ancestor.
+            logger.warning(
+                f"Could not make {parent} private ({e}); anything directly "
+                f"inside it stays readable by other users"
+            )
+            break
 
 
 def _resolve_work_dir(db_job: db.JobDB) -> Path:
--- a/fileglancer/apps/jobfiles.py
+++ b/fileglancer/apps/jobfiles.py
@@ -48,12 +48,12 @@ def ensure_private_dir(path: Path) -> None:
     0755 on a shared filesystem publishes them, and with the token the ability
     to act as that user, to everyone on the site.
 
-    Every level from ``path`` up to and including ``~/.fileglancer`` is locked
-    down, so the state root stays private however it was first created and
-    subtrees made before this was enforced are covered too. Nothing above the
-    state root is touched. An ancestor that cannot be chmod'ed (owned by another
-    uid) is warned about rather than raised, since the leaf is private either
-    way; failing to lock the leaf itself does raise.
+    Every level from ``path`` up to and including the state root is locked down,
+    so that root stays private however it was first created and subtrees made
+    before this was enforced are covered too. Nothing above it is touched. An
+    ancestor that cannot be chmod'ed (owned by another uid) is warned about
+    rather than raised, since the leaf is private either way; failing to lock
+    the leaf itself does raise.
 
     Must run as the owner of the tree (in the user worker, or as the user in CLI
     mode). Chmod rather than ``mkdir(mode=...)``: that mode is masked by the
@@ -67,9 +67,16 @@ def ensure_private_dir(path: Path) -> None:
             f"Cannot make {path} private, so it is not safe to write sensitive "
             f"Fileglancer state there: {e}"
         ) from e
+    # The state root is the innermost ``.fileglancer`` above the leaf. Found
+    # lexically, not by resolving the path: a relocated state directory (a
+    # symlinked ``~/.fileglancer``) resolves to somewhere with no
+    # ``.fileglancer`` component at all, and its real tree still wants locking.
+    state_root = next(
+        (p for p in path.parents if p.name == _STATE_DIRNAME), None
+    )
+    if state_root is None:
+        return  # not under a state directory, so there is nothing above to fix
     for parent in path.parents:
-        if _STATE_DIRNAME not in parent.parts:
-            break  # above ~/.fileglancer (or outside it): not ours to touch
         try:
             os.chmod(parent, 0o700)
         except OSError as e:
@@ -81,6 +88,8 @@ def ensure_private_dir(path: Path) -> None:
                 f"inside it stays readable by other users"
             )
             break
+        if parent == state_root:
+            break  # never ascend past the state root
 
 
 def _resolve_work_dir(db_job: db.JobDB) -> Path:
--- a/fileglancer/apps/jobfiles.py
+++ b/fileglancer/apps/jobfiles.py
@@ -64,8 +64,8 @@ def ensure_private_dir(path: Path) -> None:
         os.chmod(path, 0o700)
     except OSError as e:
         raise PermissionError(
-            f"Cannot make {path} private, so it is not safe to write job "
-            f"credentials there: {e}"
+            f"Cannot make {path} private, so it is not safe to write sensitive "
+            f"Fileglancer state there: {e}"
         ) from e
     for parent in path.parents:
         if _STATE_DIRNAME not in parent.parts:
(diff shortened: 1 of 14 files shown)


fileglancer/apps/jobfiles.py
```diff
@@ -33,6 +35,54 @@ def _build_work_dir(job_id: int, app_name: str, entry_point_id: str,
...
+    try:
+        os.chmod(path, 0o700)
+    except OSError as e:
+        raise PermissionError(
+            f"Cannot make {path} private, so it is not safe to write sensitive "
+            f"Fileglancer state there: {e}"
+        ) from e
+    for parent in path.parents:
+        if _STATE_DIRNAME not in parent.parts:
+            break  # above ~/.fileglancer (or outside it): not ours to touch
```
mkitti: This searches for `_STATE_DIRNAME` anywhere in the parent path rather than explicitly testing if the current directory is within the tree of `$HOME/.fileglancer` as the the comment states.

There is a chance this could ascend higher or lower than `$HOME/_STATE_DIRNAME`, especially if `_STATE_DIRNAME` ever became configurable to something other than `.fileglancer`.

Consider using [`is_relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) versus a fully resolved state path. This will also address paths that contain symlinks.

fileglancer/cli.py
```diff
@@ -168,6 +168,15 @@ def start(host, port, reload, workers, ssl_keyfile, ssl_certfile,
         data_dir = Path.home() / '.local' / 'share' / 'fileglancer'
         data_dir.mkdir(parents=True, exist_ok=True)
         db_path = data_dir / 'fileglancer.db'
+        # The database holds session ids and service URLs, both of which are
+        # live credentials, so it must not be readable by other users on a
+        # shared home. The directory mode is what covers SQLite's -wal and -shm
```
mkitti, line 173: Could we remove the token from the database? Do we need more than host and port?

The less we can expose there the better.

mkitti:
## Drop the token from jobs.service_url

jobs.service_url currently stores the full published URL, including the `${FG_SERVICE_TOKEN}` query string. That token is never read back from the database: the only consumer of the column is /api/apps/resolve, which calls upstream_from_service_url and keeps just the netloc. The URL returned to the browser by GET /api/jobs/{id} comes from the fresh worker read on that request, not from the column, and there is no fallback to the stored value if the worker read fails.


So the column holds a live credential for every running service with no code path that needs it. In a multi-user deployment the database is MySQL, not a file under ~, so this PR's chmod does not reach it — DB access, backups and dumps remain a way to recover every service token.

### Proposal:
In set_job_service_url (or at the call site in get_job), store only the origin — f"{scheme}://{host}:{port}" — before writing. Keep the scheme, since upstream_from_service_url relies on urlsplit and a bare host:port parses as scheme+path.
Add an Alembic data migration that rewrites existing rows the same way, so tokens already stored do not outlive the fix. Rows that fail to parse can be nulled; the next detail-page load repopulates them.
Extend the publisher/proxy tests to assert the stored value contains no query string.

This keeps /api/apps/resolve unchanged and removes the last place a service token is persisted outside the user's 0700 work directory.

— Claude Fable 5.1 Medium

fileglancer/apps/jobfiles.py
krokicki: Good catch, and the failure is concrete: given `/x/.fileglancer/y/.fileglancer/jobs/1-demo` the walk kept going past the state root and locked `/x/.fileglancer/y` and `/x/.fileglancer` too, because it only stopped once no ancestor had a `.fileglancer` component left. A home directory named `.fileglancer` would have had `$HOME` locked. Fixed in 25edee91: it now finds the innermost `.fileglancer` above the leaf and stops there, with a test for the nested case that fails without the change.

I kept it lexical rather than resolving, though, because resolving makes the relocated case worse. If someone symlinks `~/.fileglancer` at a scratch directory to keep multi-GB SIFs out of their home quota, the resolved path has no `.fileglancer` component at all, so `is_relative_to` against a resolved root would drop the walk entirely and leave the real tree unlocked -- the opposite of the intent. The lexical form follows the symlink through `chmod` and locks the target, which is what we want.

I also avoided anchoring on `Path.home() / ".fileglancer"`, since it would tie the helper to `$HOME`. `_build_work_dir` uses `expanduser(f"~{username}")` precisely because the server computes these paths on behalf of other users, and I would rather not have this one disagree with it depending on which process calls it.

If `_STATE_DIRNAME` ever does become configurable, the innermost-component search keeps working as long as the name is a single path component -- but at that point passing the root in explicitly would be better than deriving it, and I would rather make that change when there is a second value to support than guess at it now.

-- Claude Opus 5

fileglancer/cli.py
krokicki: Done in 511cf315 -- your analysis held up on every point when I traced it. The column's only reader is `/api/apps/resolve` via `upstream_from_service_url`, which keeps `urlsplit(...).netloc`; the browser's URL comes from the fresh worker read in `get_job`, and on worker failure `service_url` stays `None` with no fallback to the column. So the query string was persisted and never read.

`set_job_service_url` now stores `scheme://host:port` via a new pure `service_url_origin` in serviceproxy.py. I put it in the setter rather than the call site so the token stays out of the column whoever writes it; the import is function-local because `fileglancer.apps` imports `database`, so a module-level one would close a cycle.

One addition to the proposal: an authority carrying userinfo now stores nothing rather than the origin. That is a credential too, and `upstream_from_service_url` refuses userinfo by design, so the row would be useless anyway -- stripping it to make it dialable would have quietly widened the SSRF gate. Also worth noting `urlsplit` drops CR/LF, so a header-injection attempt no longer reaches the column at all; there is a test asserting what survives is still refused downstream.

Migration `e4d1b7a92c58` rewrites existing rows and nulls anything without a usable authority. Verified `alembic upgrade head` end to end on a scratch database, not just the helper. Tests cover the helper against the runtime one, the rewrite (token dropped, already-origin untouched, garbage and userinfo nulled), idempotency, and that a stripped row still resolves to the same upstream.

Since we are releasing from main rather than cutting 3.1.1, this rides along here instead of becoming its own PR.

-- Claude Opus 5, on behalf of @krokicki

mkitti (approved):

PR #450, opened 2026-09-10
chore: bump pypi versions

allison-truhlar:
**What changed:** pinned pypi package versions in `pyproject.toml`; `pixi.lock` file.

**Why:** the pypi package versions are pinned to prevent drift from the conda channel packages used by pixi in development. We just had a new release on production, so now is a reasonable time to update those pinned package versions. This was done using `pixi run bump-pypi-versions`.

**How this was tested:** `pixi run test-backend` - all passed. App started up and appeared to work normally with `pixi run dev-watch` and `pixi run dev-launch`. 

@krokicki @mkitti

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,9 +1,27 @@
 version: 7
 platforms:
 - name: linux-64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=x86_64
 - name: linux-aarch64
+  virtual-packages:
+  - __unix=0=0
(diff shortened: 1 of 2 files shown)


mkitti:
This seems to be working as expected and the tests do pass.

mkitti (approved):

mkitti:
There are some minor version bumps and one major version bump for x2s3. @krokicki perhaps could point us to what might be breaking in x2s3.

mkitti:
@copilot can you list what changed with the dependencies?

PR #453, opened 2026-09-16
Fix: keep host:port when a service_url has userinfo

allison-truhlar:
## What changed

`service_url_origin` (`fileglancer/apps/serviceproxy.py`) now strips only
the userinfo from a published `service_url` and keeps `scheme://host:port`,
instead of discarding the whole authority.

## Why

Commit `511cf315` added `service_url_origin` and blanked any authority
containing HTTP Basic Auth userinfo (`http://user:pass@host:port`), on the
assumption that `upstream_from_service_url` refused such URLs anyway.
That assumption held at the time — `main` had no userinfo support yet.

#448 added that support to `upstream_from_service_url`, next to `511cf315`,
without touching `service_url_origin`. The two changes never conflicted
textually, so nothing caught the mismatch: a service authenticating with
userinfo now stores nothing in `jobs.service_url`, and `/api/apps/resolve`
returns 403 for it permanently. This caused `test_publish_then_resolve_roundtrip_with_userinfo`
(`tests/test_service_proxy.py`, added by #448) to fail `403 == 204` on
`main` when #448 was merged ([failing CI run](https://github.com/JaneliaSciComp/fileglancer/actions/runs/34890903969/job/104133039739)).

`service_url_origin`'s one reader, the resolve endpoint, already strips
userinfo before using the value — it never needed the credential. Keeping
`host:port` instead of blanking the row fixes resolution without storing
the credential anywhere: the database still never sees it.

The migration `e4d1b7a92c58` inlines a copy of the same logic to stay
self-contained. `test_migration_origin_matches_runtime` enforces that the
two match, so this PR updates both.

## How this was verified

- `test_publish_then_resolve_roundtrip_with_userinfo`
  (`tests/test_service_proxy.py`, added by #448) fails `403 == 204` on
  `main` and passes with this change.
- `pixi run -e test test-backend`: 1066 passed.
- Updated three tests in `test_service_proxy_urls.py` and
  `test_service_url_migration.py` that hard-coded the old "userinfo
  discards the row" expectation, and added a test asserting the credential
  string never appears in the stored value.
- An independent review pass ran a 35-input differential sweep of
  `service_url_origin` against the migration's inlined copy (empty input,
  malformed URLs, IPv6 literals, CR/LF injection, percent-encoded `@`,
  multiple `@` characters) — zero mismatches — and confirmed the raw
  `service_url` never reaches the logs.

@krokicki


mkitti:
Is the issue that we did not test the composition of #446 and #448?

allison-truhlar:
> Is the issue that we did not test the composition of #446 and #448?

Yes, exactly that.

fileglancer/alembic/versions/e4d1b7a92c58_strip_credentials_from_job_service_urls.py
mkitti, line 30: We should better document what this function is doin.

1. This function will return `None` if
     a. The service_url is empty or otherwise `False`-like
     b. The service_url cannot be split by `urlsplit`
     c. Does not have a `scheme` or `netloc`. `netloc` includes optional credentials as well as `host:port`.
     d. If the host port portion is empty.
2. This function will return a URL as a string with just the scheme, host, and part, stripping out the path, query (with the token), fragment, and basic auth login credentials.

According to the [docs for `urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit), `urlsplit` also parses out the `username`, `password`, `host`, and `port`, so the `rpartition` may not be necessary. We could use the already split components in `parts` to parse the components, but then we would need to recombine host and port ourselves.

fileglancer/alembic/versions/e4d1b7a92c58_strip_credentials_from_job_service_urls.py
```diff
@@ -34,9 +34,12 @@ def _origin(service_url):
...
     except ValueError:
         return None
-    if not parts.scheme or not parts.netloc or '@' in parts.netloc:
+    if not parts.scheme or not parts.netloc:
         return None
-    return urlunsplit((parts.scheme, parts.netloc, '', '', ''))
+    _userinfo, _sep, hostport = parts.netloc.rpartition('@')
+    if not hostport:
+        return None
+    return urlunsplit((parts.scheme, hostport, '', '', ''))
```
mkitti, line 42: Here the five components are

1. scheme
2. netloc
3. path (empty)
4. query (empty)
5. fragment (empty)

Rather than the tuple, perhaps we should `urllib.parse.SplitResult` with the arguments passed as keyword arguments.

fileglancer/apps/serviceproxy.py
```diff
@@ -142,21 +142,30 @@ def service_url_origin(service_url: Optional[str]) -> Optional[str]:
...
         return None
     try:
         parts = urlsplit(service_url)
     except ValueError:
         return None
-    if not parts.scheme or not parts.netloc or '@' in parts.netloc:
+    if not parts.scheme or not parts.netloc:
         return None
-    return urlunsplit((parts.scheme, parts.netloc, '', '', ''))
+    _userinfo, hostport = _split_userinfo(parts.netloc)
```
mkitti, line 165: I think we should consider refactoring this as well to lean more on `urlsplit` rather than having a separate function to split the userinfo off.

mkitti (approved):
I was considering that we should do a refactor in this pull request, but I think it might be better to merge this as is and refactor in another pull request.

allison-truhlar:
> I was considering that we should do a refactor in this pull request, but I think it might be better to merge this as is and refactor in another pull request.

Sounds good - I will merge as-is. I added a ClickUp task to refactor, with a link to the comments here.

